PR #1, opened 2017-03-30
244 vertical slice

scoch:
Functionality so far for first vertical slice. Includes service and unit tests.

Things to especially critique:
* structure
* tests
* approach to service


app/views/index.html
```diff
@@ -16,7 +16,8 @@ <h1 class="heading-xlarge">
       </h1>
 
       <ul>
-          <li><a href="/example">Example Link</a></li>
+        <li><a href="/example">Example Link</a></li>
```
andrew-js-wright: Can probably remove this now.

gulpfile.js
```diff
@@ -62,8 +64,27 @@ gulp.task('sass', function () {
...
+gulp.task('generate-assets-and-start', ['generate-assets', 'server'], function () {
+
+})
+
+gulp.task('nsp', function (cb) {
+  gulpNSP({package: __dirname + '/package.json'}, cb);
+})
+
+gulp.task('watch', function() {
+  gulp.watch(['app/app.js', 'app/services/**/*'], ['generate-assets-and-start']);
```
andrew-js-wright: 😍 can we also track `app/assets/**/*` to pick up sass and frontend Javascript changes?

app/views/index.html
```diff
@@ -16,7 +16,8 @@ <h1 class="heading-xlarge">
...
+        <li><a href="/my-utilisation">My Utilisation</a></li>
```
andrew-js-wright: href needs update to reflect additional URL parameters.

app/routes/offender-manager-utilisation.js
```diff
@@ -0,0 +1,11 @@
+const ViewUtilisation = require('../services/domain/view-utilisation')
+
+module.exports = function (router) {
+  router.get('/offender-manager-utilisation/:userId/:year/', function (req, res, next) {
```
tswann: User Id implies it's a user of the system which won't always be the case. `offenderManagerId` would probably be more accurate and reflective of how it'll be named in the DB

app/views/index.html
```diff
@@ -16,8 +16,7 @@ <h1 class="heading-xlarge">
       </h1>
 
       <ul>
-        <li><a href="/example">Example Link</a></li>
-        <li><a href="/my-utilisation">My Utilisation</a></li>
+        <li><a href="/offender-manager-utilisation/1/2017">My Utilisation</a></li>
```
tswann: I think this should be called `caseload-utilisation` as although the current scope is to filter for an individual Offender Manager it will be expanded to include filters for Team, LDU etc.

See https://noms-wmt-ux.herokuapp.com/alpha/v3-2/capacity

This also applies to the 'past'/'current' description in the current taiga stories I think - it's referring to the filters on the same view effectively.

app/routes/offender-manager-utilisation.js
```diff
@@ -0,0 +1,9 @@
+const getOffenderManagerUtilisationTable = require('../services/get-offender-manager-utilisation-table')
+
+module.exports = function (router) {
+  router.get('/offender-manager-utilisation/:userId/:year/', function (req, res, next) {
```
tswann: I think this should be `caseload-utilisation` as the same view should ultimately be extended to include filters for different levels of Org unit. I.e. LDU, Team etc.

As per https://noms-wmt-ux.herokuapp.com/alpha/v3-2/capacity in the prototype

`userId` also implies it's always a user of the system, which probably isn't the case. 

What about something like `/caseload-utilisation/offendermanager/:id/:year/`

I'm thinking this would allow us to evolve the page for different levels of org unit by just changing the `offendermanager` bit for the appropriate category. What do you think?

app/services/data/get-offender-manager-utilisation.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function (userId, year) {
```
tswann: Just extending the comment above - if we had this be `(orgunittype, id, year)` we can start with offender manager and extend to include the others later.

PR #5, opened 2017-04-18
Feature/wmt 243/utilisation by year

scoch:
Added date pickers
Updated mock data to reflect on a per day capacity vs per month.
Updated plot.ly to account for the new data format on a basic level
Added validation checks on domain object utilisation-date
Added frontend validation error header message
Added tests for validation
Updated existing tests that have changed.


app/services/validators/common-validator.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+exports.isDateInThePast = function (date) {
+  return this.isValidDate(date) &&
+    date < dateFormatter.now()
+}
+
+exports.isDateOlderThanMaxHistory = function (date) {
+  var yearsAgo = dateFormatter.now().diff(date, 'years')
+  return yearsAgo > 6
```
andrew-js-wright: Would be good to pull this out into a const somewhere that can be shared across the app.

test/unit/services/test-date-formatter.js
```diff
@@ -0,0 +1,50 @@
...
+const bases = require('bases')
+const DATE_FORMAT = 'YYYY-MM-DD'
+const INVALID_DATE_ERROR = 'Invalid date'
+
+describe('services/date-formatter', function () {
+  const VALID_DAY = '01'
+  const VALID_MONTH = '01'
+  const VALID_YEAR = '2000'
+  const EXPECTED_DATE = moment([VALID_YEAR, VALID_MONTH, VALID_DAY], DATE_FORMAT)
+  const ENCODED_DATE = bases.toBase8(EXPECTED_DATE.format('YYYYMMDD'))
```
scoch: Cheers! Removed unused refs and dependency

app/views/utilisation.html
```diff
@@ -10,12 +10,58 @@ <h2 class="heading-large">
...
+    <div class="form-group">
+        {% include "includes/validation-error-messages.html" %}
+        <fieldset>
```
scoch: Wanted to ask this one. Do we want to validate the to/from fields individually or as one collective 'date range'? I'll pop this in the slack too

app/services/domain/utilisation-date.js
```diff
@@ -0,0 +1,35 @@
...
+    ]
+
+    this.utilisationDate = dateFormatter.build(day, month, year)
+    this.IsValid()
+  }
+
+  IsValid () {
+    var errors = ErrorHandler()
+
+    FieldsetValidator(this.fields, 'utilisationDate', errors)
```
scoch: Okay cool. Can change this based on the change from above. Might need to tag where the data is coming from in this case as so far it just treats it as two of the same type of dates.

app/routes/offender-manager-utilisation.js
```diff
@@ -1,10 +1,55 @@
...
+        toDate = dateFormatter.now()
+
+        fromUtilisationDate = new UtilisationDate(
+          fromDate.date(),
+          fromDate.month() +1,
+          fromDate.year()
+        )
+        toUtilisationDate = new UtilisationDate(
+          toDate.date(),
+          toDate.month() +1,
```
scoch: moment has a zero starting index for months which can throw it off a bit (this means 3 === April). I can move this to the object itself but I wasn't sure if hiding this quirk of the moment lib was wise or not. 

Also .day() caught me out too, returning the day of the week. Thats why it's using .date()

app/views/utilisation.html
andrew-js-wright: I think, for now, treat them as individual dates as we might change this based on user research discussions.

app/services/domain/utilisation-date.js
andrew-js-wright: Cool, might be worth passing the field name into the validator.

app/routes/offender-manager-utilisation.js
```diff
@@ -1,10 +1,50 @@
...
+        utilisation: getUtilisationTable(ORG_UNIT_TYPE.OFFENDER_MANAGER, req.params.id, capacityDateRange)
+      })
+
+    } catch (error) {
+      if (error instanceof ValidationError) {
+        return res.status(400).render('utilisation', {
+          errors: error.validationErrors
+        })
+      } else {
+        throw error
```
dconey646: What does the application do at this point?

dconey646:
Just one comment, but apart from that, looks good to me!

app/routes/offender-manager-utilisation.js
scoch: Ripped this from APVS approach, from what I understand this is an error outside of validation scope so it'll be thrown and handled by the error handling in app.js. This then renders the error.html page. Haven't looked into that scenario much yet. Might be safer waiting until we have the backend hooked up and test a different error.

dconey646: Ah, great. Wasn't too sure if this was a random error thrown. Good to see it handled by the app.js.

test/unit/services/validators/test-common-validator.js
```diff
@@ -0,0 +1,129 @@
...
+    })
+
+    it('should return false if passed an invalid Date object', function () {
+      var result = validator.isDateInThePast(FUTURE_DATE)
+      expect(result).to.equal(false)
+    })
+  })
+
+  describe('isDateOlderThanMaxHistory', function () {
+    const MAX_HISTORY = 6
```
andrew-js-wright: Now that we have this in a constant we can include it so that future updates only require changing one place.

test/unit/services/validators/test-fieldset-validator.js
```diff
@@ -0,0 +1,160 @@
...
+    })
+
+    it('should return the fieldsetValidator after being called to allow function chaining.', function () {
+      var result = this.fieldsetValidator.isPastDate(dateFormatter.now())
+      expect(result).to.be.equal(this.fieldsetValidator)
+    })
+  })
+
+  describe('isOlderThanMaxHistory', function () {
+    const MAX_HISTORY = 6
```
andrew-js-wright: This can be retrieved from the constant too.

PR #6, opened 2017-04-25
WMT-275: corrected standardjs config and any errors that were present…

scoch:
…. Only errors remain are to do with test data.


andrew-js-wright:
Sigh didn't use the correct formatting on that last comment and the mobile site won't let me edit.

andrew-js-wright:
My bad, it's there already. Cheers Steven!

gulpfile.js
```diff
@@ -18,7 +18,13 @@ gulp.task('assets', function () {
 })
 
 gulp.task('standard', function () {
-   return gulp.src(['app/app.js'])
+  return gulp.src([
+    '!node_modules/*',
+    '!app/govuk_modules/**/*.js',
+    '!app/assets/javascripts/**/*.js',
+    '!app/public/javascripts/**/*.js',
+    'app/**/*.js'
```
andrew-js-wright: I would also include `app/*.js` to catch files such as `app/app.js`

PR #10, opened 2017-05-30
417: Query the database for capacity percentage

andrew-js-wright:
Add integration tests too


andrew-js-wright:
Need to split out the integration and unit test runs and update the travis file before this builds green.

test/helpers/data/workload-capacity-helper.js
```diff
@@ -0,0 +1,132 @@
+const config = require('../../../knexfile')
+const knex = require('knex')(config)
+var Promise = require('bluebird').Promise
+
+module.exports.addWorkloadCapacitiesForOffenderManager = function () {
+  var inserts = []
+
+  return knex('offender_manager_type').returning('id').insert({description: 'test'})
```
scoch: personal pref, but what about having returning at the end as it'd be the last action of the call chain.

test/integration/services/data/get-capacity-for-individual.js
```diff
@@ -0,0 +1,36 @@
...
+describe('services/data/get-capacity-for-individual', function () {
+  before(function (done) {
+    workloadCapactiyHelper.addWorkloadCapacitiesForOffenderManager()
+      .then(function (builtInserts) {
+        inserts = builtInserts
+        done()
+      })
+  })
+
+  it('should retrieve all the workloads withiin the date range', function (done) {
```
scoch: withiin

Apologies for the grammar police 😔

PR #12, opened 2017-06-08
Stop app crashing when worklaod user not found

andrew-js-wright:
Currently, when the workload user is undefined, the application crashes because
the `DisplayTable` object is validating it has more than one header.

For now, the expected behaviour should be displaying an empty graph the users
with default headings.

In the future we'll want to handle this with a 404 response before the
`get-capacity-table` function is called.


scoch:
+1

PR #14, opened 2017-06-23
Feature/wmt 562/fix graph

dconey646:


app/assets/javascripts/caseload-capacity.js
```diff
@@ -67,6 +77,15 @@ layout = {
...
+    },
+    yaxis: {
+        autorange: false,
+        fixedrange: true,
+        range: [0, 120],
+        showline: true,
+        ticks: 'outside',
+        title: 'Capacity %',
+        type: 'linear'
     }
```
andrew-js-wright: This `yaxis` object is not required as it will overwrite the previous one.

app/assets/javascripts/caseload-capacity.js
```diff
@@ -9,7 +9,17 @@ trace1 = {
...
-data = [trace1];
+trace2 = {
+    x: capacityTable.headings,
+    y: capacityTable.rows[1].values,
+    line: {width: 3.5},
+    marker: {
+        line: {width: -0.5},
+        size: 9
+    },
+    name: 'Reduction Hours ',
```
andrew-js-wright: Trailing space on `Reduction Hours`.

PR #21, opened 2017-07-11
334: Add sub-navigation bar

andrew-js-wright:
A navigation item is very similar to a breadcrumb item. Therefore refactor the
Breadcrumb object to a Link object and use for both breadcrumbs and sub-nav.

Import the sub nav sass from the ux prototype, refactoring slightly to avoid
duplication.

Extend the link generator in order to allow link generation from tree references
or name and id references. This allows us to reuse the link generation
functionality in the sub nav builder.

Build the sub-navigation at the same time as we're building the breadcrumbs and
the title/sub-title, i.e. in the get-capacity-view service. This will have to be
done on all requests and therefore should be extracted out to reduce
duplication.

screen shot 2017-07-11 at 10 18 09


test/unit/services/test-get-capacity-view.js
```diff
@@ -136,73 +136,63 @@ const CAPACITY_RESULTS = [
...
+var getCapacityView
+var expectedTitle = breadcrumbHelper.LDU_BREADCRUMBS[0].title
+
+before(function () {
+  getCapacityStub = sinon.stub()
+  getBreadcrumbs = sinon.stub().returns(breadcrumbHelper.LDU_BREADCRUMBS)
+  getCapacityView = proxyquire('../../../app/services/get-capacity-view',
+    {'./data/get-individual-workload-reports': getCapacityStub,
+      './data/get-workload-report-views': getCapacityStub,
+      './get-breadcrumbs': getBreadcrumbs})
```
caoimheb: Stub needed for `getSubNav`?

test/unit/services/test-link-generator.js
```diff
@@ -3,27 +3,27 @@ const linkGenerator = require('../../../app/services/helpers/link-generator')
...
   it('should throw an error when passed a reference of an incorrect format', function () {
-    expect(() => linkGenerator('123')).to.throw(/incorrect format/)
-    expect(() => linkGenerator('r1')).to.throw(/incorrect format/)
-    expect(() => linkGenerator('Region1')).to.throw(/incorrect format/)
-    expect(() => linkGenerator('Team1A')).to.throw(/incorrect format/)
+    expect(() => linkGenerator.fromReference('123')).to.throw(/incorrect format/)
+    expect(() => linkGenerator.fromReference('r1')).to.throw(/incorrect format/)
+    expect(() => linkGenerator.fromReference('Region1')).to.throw(/incorrect format/)
+    expect(() => linkGenerator.fromReference('Team1A')).to.throw(/incorrect format/)
   })
```
caoimheb: Do you need to include tests for `fromIdAndName()` explicitly?

test/unit/services/test-get-capacity-view.js
andrew-js-wright: ✅

test/unit/services/test-link-generator.js
andrew-js-wright: ✅

andrew-js-wright:
Thanks for the review @caoimheb, suggestions implemented.

PR #22, opened 2017-07-11
Feature/wmt 334/view individual case progress

dconey646:


test/helpers/data/workload-capacity-helper.js
```diff
@@ -53,8 +53,9 @@ module.exports.addWorkloadCapacitiesForOffenderManager = function () {
       inserts.push({table: 'workload', id: ids[0]})
       return knex('workload_report').returning('id').insert([
         { effective_from: new Date(2008, 0, 1) },
-        { effective_from: new Date(2009, 0, 1) },
-        { effective_from: new Date(2011, 0, 1) }
+        { effective_from: new Date(2009, 0, 1), effective_to: new Date(2010, 0, 1) }
+        // { effective_from: new Date(2011, 0, 1), effective_to: new Date(2010, 0, 1) },
```
dconey646: Forgot I left those in, will remove them after discussion.

test/helpers/data/workload-capacity-helper.js
```diff
@@ -135,6 +130,33 @@ module.exports.addWorkloadCapacitiesForOffenderManager = function () {
...
+
+      for (let i = 1; i <= 7; i++) {
+        defaultTier.tier_number = i
+        defaultTier.location = 'COMMUNITY'
+        tiers.push(defaultTier)
+        defaultTier.location = 'LICENSE'
+        tiers.push(defaultTier)
+        defaultTier.location = 'CUSTODY'
+        tiers.push(defaultTier)
+      }
```
dconey646: Need to change this as well, marking as a note.

test/integration/services/data/test-get-individual-caseload-progress.js
```diff
@@ -0,0 +1,40 @@
...
+  before(function (done) {
+    workloadCapacityHelper.addWorkloadCapacitiesForOffenderManager()
+      .then(function (builtInserts) {
+        inserts = builtInserts
+        done()
+      })
+  })
+
+  it('should retrieve current caseload progress for an offender manager', function (done) {
+    // TODO: Write test to get individual caseload progress and verify
```
dconey646: GOing to remove

test/integration/services/data/test-get-individual-caseload-progress.js
```diff
@@ -0,0 +1,39 @@
...
+  before(function (done) {
+    workloadCapacityHelper.addWorkloadCapacitiesForOffenderManager()
+      .then(function (builtInserts) {
+        inserts = builtInserts
+        done()
+      })
+  })
+
+  it('should retrieve current caseload progress for an offender manager', function (done) {
+    getIndividualCaseloadProgress(inserts.filter((item) => item.table === 'offender_manager')[0].id)
```
andrew-js-wright: This should be `workload_owner` table's id.

test/helpers/data/workload-capacity-helper.js
```diff
@@ -135,6 +128,29 @@ module.exports.addWorkloadCapacitiesForOffenderManager = function () {
...
+        total_cases: 10,
+        location: 'COMMUNITY'}
+
+      var tiers = []
+
+      for (let i = 1; i <= 7; i++) {
+        tiers.push(Object.assign({}, defaultTier, {tier_number: i, location: 'COMMUNITY'}))
+        tiers.push(Object.assign({}, defaultTier, {tier_number: i, location: 'CUSTODY'}))
+        tiers.push(Object.assign({}, defaultTier, {tier_number: i, location: 'LICENSE'}))
+      }
```
andrew-js-wright: I'm not seeing the value of adding so many tiers. I know it's more like the real data but I don't think we need it all to prove that query works. We could probably get away with just 2 tier rows and verify that the values we expect are summed correctly. It would keep things similar in the helper and reduce database load.

app/services/data/get-individual-caseload-progress.js
```diff
@@ -0,0 +1,22 @@
...
+    .where('workload_owner_id', id)
+    .whereNotNull('workload_report.effective_from')
+    .whereNull('workload_report.effective_to')
+    .max('workload.community_last_16_weeks AS community_last_16_weeks')
+    .max('workload.license_last_16_weeks AS license_last_16_weeks')
+    .max('workload.total_cases AS total_cases')
+    .sum('tiers.warrants_total AS warrants_total')
+    .sum('tiers.overdue_terminations_total AS overdue_terminations_total')
+    .sum('tiers.unpaid_work_total AS unpaid_work_total')
+    .groupBy('workload.id')
```
andrew-js-wright: Can we add a `first()` to this so that the DB only returns one record (as that's all we expect)

app/services/data/get-individual-caseload-progress.js
```diff
@@ -0,0 +1,22 @@
...
+    .leftJoin('tiers', 'workload.id', 'tiers.workload_id')
+    .where('workload_owner_id', id)
+    .whereNotNull('workload_report.effective_from')
+    .whereNull('workload_report.effective_to')
+    .max('workload.community_last_16_weeks AS community_last_16_weeks')
+    .max('workload.license_last_16_weeks AS license_last_16_weeks')
+    .max('workload.total_cases AS total_cases')
+    .sum('tiers.warrants_total AS warrants_total')
+    .sum('tiers.overdue_terminations_total AS overdue_terminations_total')
+    .sum('tiers.unpaid_work_total AS unpaid_work_total')
```
andrew-js-wright: Can we update the `AS` aliases so the fields are returned in camel case instead of snake case i.e. `unpaidWorkTotal` etc.

test/integration/services/data/test-get-individual-caseload-progress.js
dconey646: Yep, good spot.

test/helpers/data/workload-capacity-helper.js
dconey646: My thinking was to be more like real data, but I am happy to change it to less data.

app/services/data/get-individual-caseload-progress.js
dconey646: Good idea, done.

app/services/data/get-individual-caseload-progress.js
dconey646: Yep, done.

PR #23, opened 2017-07-11
Feature/334/335/view case progress chart

caoimheb:


andrew-js-wright:
some standard issues being flagged by travis.

caoimheb:
Travis error `Error: Cannot find module './data/get-individual-caseload-progress'`. Module is placeholder for DB query file in other branch

app/views/case-progress.html
```diff
@@ -0,0 +1,51 @@
...
+{% include "includes/breadcrumbs.html" %}
+
+{% include "includes/validation-error-messages.html" %}
+
+<h2 class="heading-large">
+    <span class="heading-secondary">
+        {{ subTitle | title }}
+    </span>
+    {{ title | title }}
+</h2>
```
andrew-js-wright: A nice improvement here might be to extract this into a partial like the breadcrumbs, as we're now using it multiple places, and will likely need it on most pages to come.

caoimheb: Good shout. Updated

app/routes/capacity-route.js
```diff
@@ -14,7 +14,7 @@ module.exports = function (router) {
       return res.render('capacity', {
         title: result.title,
         subTitle: result.subTitle,
-        subnav: getSubNav(id, organisationLevel, req.path),
+        subNav: getSubNav(id, organisationLevel, req.path),
```
andrew-js-wright: 👍

PR #24, opened 2017-07-11
334: Call getSubNav from the route instead of the service

andrew-js-wright:
The getSubNav call is going to be common across all of our screens. So it should
be as self contained as possible.

In stead of building up a list of navigation items in the service and then
setting the active flag on one, based on the req.path parameter, in the route.
The subNav array, including the active flag, is built in the get-sub-nav
service.

The service is then called only in the route, where we have access to the path.
This reduces the footprint of this common functionality and will make it easier
to bring new routes on board.

### DIFF
--- a/app/routes/capacity-route.js
+++ b/app/routes/capacity-route.js
@@ -1,6 +1,6 @@
 const getCapacityView = require('../services/get-capacity-view')
 const dateRangeHelper = require('../services/helpers/date-range-helper')
-const _ = require('lodash')
+const getSubNav = require('../services/get-sub-nav')
 
 module.exports = function (router) {
   router.get(`/:organisationLevel/:id/caseload-capacity`, function (req, res, next) {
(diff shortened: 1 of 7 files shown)


app/routes/capacity-route.js
```diff
@@ -11,12 +11,10 @@ module.exports = function (router) {
     var capacityViewPromise = getCapacityView(id, capacityDateRange, organisationLevel)
 
     return capacityViewPromise.then(function (result) {
-      var activeIndex = _.findIndex(result.subNav, {link: req.path})
-      result.subNav[activeIndex].active = true
       return res.render('capacity', {
         title: result.title,
         subTitle: result.subTitle,
-        subnav: result.subNav,
+        subnav: getSubNav(id, organisationLevel, req.path),
```
caoimheb: `subNav` as opposed to `subnav`, for consistency?

andrew-js-wright: Good catch. I'll let your branch over rule this laziness.

PR #33, opened 2017-07-18
698: Add national level capacity view

andrew-js-wright:
The caseload-capacity screen should be visable at a national level. This will
display the aggregated capacity for every offender manager in the system. As
there is only one instance of the national level, no id field is required.
However a dummy id `0` is used in order to keep the routing consistent.

Therefore the national level capacity screen is accessed from the path:

`/nps/0/caseload-capacity`

### DIFF
--- a/app/services/get-breadcrumbs.js
+++ b/app/services/get-breadcrumbs.js
@@ -20,6 +20,7 @@ module.exports = function (id, organisationLevel) {
     breadcrumbs.push(new Link(tree[reference].name, linkGenerator.fromReference(reference)))
     reference = tree[reference].parent
   } while (reference !== undefined)
+
   return breadcrumbs
 }
 
--- a/app/services/get-capacity-view.js
+++ b/app/services/get-capacity-view.js
@@ -11,7 +11,7 @@ module.exports = function (id, capacityDateRange, organisationLevel) {
   var workloadReportsPromise
 
   if (organisationalUnitType === undefined) {
-    throw new Error(organisationLevel + ' should be offender-manager, region, team or ldu')
+    throw new Error(organisationLevel + ' should be offender-manager, region, team, ldu or hmpps')
   }
 
   if (organisationLevel === routeType.OFFENDER_MANAGER.name) {
--- a/app/views/includes/breadcrumbs.html
+++ b/app/views/includes/breadcrumbs.html
@@ -1,4 +1,5 @@
 <div class="breadcrumbs">
+{% if breadcrumbs.length > 1 %}
   <ol>
   {% for breadcrumb in breadcrumbs.reverse() %}
     {% if loop.last %}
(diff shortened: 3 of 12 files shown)


andrew-js-wright:
The caseload-capacity integration tests will need to be extended in order to ensure national level view is working correctly. However, it would make most sense to wait until @caoimheb has created the test data for case-progress as this should be reusable here.

app/services/get-breadcrumbs.js
```diff
@@ -20,6 +20,11 @@ module.exports = function (id, organisationLevel) {
     breadcrumbs.push(new Link(tree[reference].name, linkGenerator.fromReference(reference)))
     reference = tree[reference].parent
   } while (reference !== undefined)
+
+  if (breadcrumbs.length < 2) {
+    breadcrumbs = []
+  }
```
dconey646: Whats the reasoning behind this? Is this due to National level being the only breadcrumb?

app/services/get-capacity-view.js
```diff
@@ -11,7 +11,7 @@ module.exports = function (id, capacityDateRange, organisationLevel) {
   var workloadReportsPromise
 
   if (organisationalUnitType === undefined) {
-    throw new Error(organisationLevel + ' should be offender-manager, region, team or ldu')
+    throw new Error(organisationLevel + ' should be offender-manager, region, team, ldu or nps')
   }
```
caoimheb: Did Simon mention that we should stop using nps? Think he said switch to hmpps? 

Maybe a separate branch to refactor that?

app/services/get-breadcrumbs.js
andrew-js-wright: Yea, there was no need to have a label saying `NPS` when there was no possible interaction with the breadcrumbs.

dconey646: Sounds good to me.

app/services/get-capacity-view.js
andrew-js-wright: After chatting to Simon, I think hmpps is the way to go. Good shout Caoimhe.

app/views/includes/breadcrumbs.html
```diff
@@ -10,4 +11,5 @@
       <!-- <li>This would display if the breadcrumbs collection were empty</li> -->
```
caoimheb: This else can now be removed?

test/unit/services/test-get-breadcrumbs.js
```diff
@@ -52,8 +52,8 @@ describe('services/get-breadcrumbs', function () {
     }
     var getBreadcrumbs = proxyquire('../../../app/services/get-breadcrumbs', {'./organisational-hierarchy-tree': getTree})
 
-    expect(getBreadcrumbs(1, 'nps').length).to.eql(1)
-    expect(getBreadcrumbs(1, 'nps')).to.eql(breadcrumbDataHelper.NATIONAL_BREADCRUMBS)
+    expect(getBreadcrumbs(1, 'hmpps').length).to.eql(0)
```
caoimheb: Test failing

andrew-js-wright:
Applying comments

PR #34, opened 2017-07-18
Feature/wmt 636/case progress views

caoimheb:


test/helpers/data/workload-capacity-helper.js
```diff
@@ -2,109 +2,170 @@ const config = require('../../../knexfile').integrationTests
...
+    return addLdu(inserts)
+  })
+  .then(function (inserts) {
+    return addTeam(inserts)
+  })
+  .then(function (inserts) {
+    return addOffenderManager(inserts)
+  })
+  .then(function (inserts) {
+    return addPopulatedRegion(inserts)
```
andrew-js-wright: Can we apply this pattern of adding the children when we add a parent organisation?

test/integration/services/data/test-get-caseload-progress.js
```diff
@@ -0,0 +1,176 @@
...
+
+var omCaseProgressRow = {
+  name: 'Test_Forename Test_Surname',
+  communityLast16Weeks: 10,
+  licenseLast16Weeks: 9,
+  totalCases: 5,
+  warrantsTotal: 30,
+  overdueTerminationsTotal: 30,
+  unpaidWorkTotal: 30
+}
```
andrew-js-wright: Replace these with generator method, as discussed.

test/helpers/data/workload-capacity-helper.js
caoimheb: Updated

test/integration/services/data/test-get-caseload-progress.js
caoimheb: Added

PR #35, opened 2017-07-18
Feature/wmt 636/view case progress bar chart updates

dconey646:


app/assets/javascripts/case-progress.js
```diff
@@ -1,65 +1,83 @@
...
-  x: [caseProgress.communityLast16Weeks], 
-  y: [rowTitle], 
+  x: commLast16WeeksX,
+  y: namesY,
   hoverinfo: 'x', 
   marker: {color: 'rgb(131, 202, 207)'}, 
   name: 'Community <16WK', 
   orientation: 'h', 
   type: 'bar', 
   uid: 'ccd732'
```
caoimheb: Are these uids needed? I meant to remove them in individual, but forgot....

app/services/get-case-progress.js
```diff
@@ -1,14 +1,14 @@
...
 
 module.exports = function (id, organisationLevel) {
   var result = {}
   var organisationalUnitType = getOrganisationUnit('name', organisationLevel)
-  var caseProgressPromise = getCaseProgress(id)
+  var caseProgressPromise = getCaseProgress(id, organisationLevel)
 
   return caseProgressPromise.then(function (results) {
-    result.caseProgress = results
+    result.caseProgressList = [results]
```
caoimheb: Does this need to be wrapped in `[]`? It will now be returning an Array, so can the Array itself not be used?

app/views/case-progress.html
```diff
@@ -14,21 +14,25 @@
...
         <th>Community Last 16 Weeks</th>
         <th>License Last 16 Weeks</th>
         <th>Total Cases</th>
         <th>Warrants Total</th>
         <th>Unpaid Work Total</th>
         <th>Overdue Terminations Total</th>
     </tr>
+    {% for caseProgress in caseProgressList%}
     <tr>
+        <td>{{ caseProgress.name }}</td>
```
andrew-js-wright: This should be a `<th>` because it's a header I think. Perhaps @stevenadams could shed some light on this?

app/assets/javascripts/case-progress.js
dconey646: Not 100% sure, Could confirm with @stevenadams.

PR #44, opened 2017-07-21
Feature/wmt 678/overview front end

caoimheb:


caoimheb:
Travis will fail as `get-overview` is calling a query which does not yet exist

caoimheb:
@andrew-js-wright Some new logic added to include capacity percentage.

andrew-js-wright:
There seems to be an empty file `app/services/data/get-individual-overview.js`

PR #49, opened 2017-07-25
678: Add query for individual overview details

caoimheb:
Update data helper used in integration tests to include data needed
to test individual overview

### DIFF
--- a/app/services/data/get-individual-overview.js
+++ b/app/services/data/get-individual-overview.js
@@ -1,0 +1,24 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+
+module.exports = function (id) {
+  return knex('workload_owner')
+    .join('team', 'workload_owner.team_id', 'team.id')
+    .join('workload', 'workload.workload_owner_id', 'workload_owner.id')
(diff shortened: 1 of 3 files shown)


app/services/data/get-individual-overview.js
```diff
@@ -0,0 +1,24 @@
...
+  return knex('workload_owner')
+    .join('team', 'workload_owner.team_id', 'team.id')
+    .join('workload', 'workload.workload_owner_id', 'workload_owner.id')
+    .join('workload_points_calculations', 'workload_points_calculations.workload_id', 'workload.id')
+    .join('offender_manager', 'offender_manager.id', 'workload_owner.offender_manager_id')
+    .first('offender_manager.grade_code AS grade',
+           'team.id AS teamId',
+           'team.description AS teamName',
+           'workload_points_calculations.available_points AS availablePoints',
+           'workload_points_calculations.total_points AS totalPoints',
```
andrew-js-wright: Unless the total_points field has been updated in the database to do this already, we're going to need to extract and sum all of the fields required to get the actual total points i.e.: `workload points + sdr points + sdr conversion points + PAROMS points`

caoimheb: PR now in to update this in the worker to sum all four values and store them in total_points

PR #50, opened 2017-07-25
Replace total_points sum with total_points value in capacity-calculator

caoimheb:


app/services/helpers/capacity-calculator.js
```diff
@@ -2,9 +2,7 @@
 module.exports.calculate = function (workloadReports) {
   var capacities = []
   workloadReports.forEach(function (report) {
-    var allPoints = report.total_points + report.sdr_points +
-      report.sdr_conversion_points + report.paroms_points
-    var capacityPercentage = (allPoints / report.available_points) * 100
+    var capacityPercentage = (report.total_points / report.available_points) * 100
```
andrew-js-wright: Should we also update the query to not return `srd_points`, `sdr_conversion_points` and `paroms_points`?

caoimheb: Removed in `wmt-web` and `wmt-worker` on the `update_total_points` branches

PR #51, opened 2017-07-25
Refactor/pa11y errors

caoimheb:


run-pa11y.sh
```diff
@@ -15,19 +15,21 @@ lduId=$2
 teamId=$3
 omId=$4
 
+pa11yCommand="""pa11y --standard WCAG2AA --ignore "notice" --hide-elements "div[role=presentation]" --reporter html"""
```
andrew-js-wright: 👍

caoimheb:
@andrew-js-wright @tswann Updates to the script for use in Jenkins

caoimheb:
TODO: Add overview tests

run-pa11y.sh
```diff
@@ -15,19 +14,44 @@ lduId=$2
 teamId=$3
 omId=$4
 
+pa11yCommand="""pa11y --standard WCAG2AA --ignore "warning\;notice" --hide-elements "div[role=presentation]" """
+host="http://localhost:3000"
```
andrew-js-wright: Can we pull the value of the base path from configuration so it can be run independently on each environment?

run-pa11y.sh
```diff
@@ -15,19 +14,44 @@ lduId=$2
...
+urls+=("$host/region/$regionId/$case_progress_url")
+
+# Overview
+overview_url='overview'
+urls+=("$host/offender-manager/$omId/$overview_url")
+
+for url in "${urls[@]}"
+do
+  $pa11yCommand $url
+  if [ $? == 2 ]
```
andrew-js-wright: Can we just check if the error code is non-zero. That way the Jenkins job will fail, both when there is an accessibility error and when the tool just didn't run for some other reason.

caoimheb: Updated

run-pa11y.sh
caoimheb: Updated

andrew-js-wright:
Looks good, thanks @caoimheb. Can we add a node script that used the DB environment variables to go, grab existing workload owner / team / ldu / region ids, and pass them to pa11y so we can run this on Jenkins.

PR #62, opened 2017-08-01
Fix/e2e tests

andrew-js-wright:


test/helpers/data/aggregated-data-helper.js
```diff
@@ -144,17 +135,24 @@ var addOffenderManager = function (inserts) {
...
     .then(function (ids) {
-      inserts.push({table: 'workload', id: ids[0]})
-
-      var workloadReports = inserts.filter((item) => item.table === 'workload_report')
+      ids.forEach((id) => {
+        inserts.push({table: 'workload', id: id})
+      })
 
+      return knex('workload_report').select('id').orderBy('effective_from', 'desc')
+    }).then(function (workloadReports) {
```
caoimheb: Are you returning workload reports here, or just the ids?

test/e2e/view-offender-manager-overview.js
```diff
@@ -9,7 +9,8 @@ describe('View overview', function () {
   before(function () {
     return dataHelper.selectIdsForWorkloadOwner()
       .then(function (results) {
-        workloadOwnerId = results.filter((item) => item.table === 'workload_owner')[0].id
+        var workloadOwnerIds = results.filter((item) => item.table === 'workload_owner')
+        workloadOwnerId = workloadOwnerIds[workloadOwnerIds.length - 1].id
```
dconey646: Whats the thinking behind getting the last workload owner id?

test/helpers/data/aggregated-data-helper.js
andrew-js-wright: It is just the id but if we were to name it `workloadReportIds` then in order to access the first id we would still have to reference `workloadReportIds[0].id`. This is because the object actually does represent a workload report object, even though, in this instance, we are only returning the ID component. After a conversation about it, we've agreed to leave it as it is.

test/e2e/view-offender-manager-overview.js
andrew-js-wright: It actually doesn't matter, there should only be one. Nice catch, I'll revert because it's confusing.

PR #63, opened 2017-08-01
680: Refactoring task of ensuring we get the latest contracted hours.…

dconey646:
… Also introducing new table layer for the frontend. Query to follow


app/services/data/get-team-caseload-overview.js
```diff
@@ -0,0 +1,18 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+
+module.exports = function (id) {
+  return knex('app.team_case_overview')
```
andrew-js-wright: Schema name is not required here.

app/views/includes/team-overview-table.html
```diff
@@ -0,0 +1,35 @@
...
+      <tr >
+        <td headers="name om"><a href="/offender-manager/{{ member.personId }}">{{ member.name }}</td>
+        <td headers="grade om">{{ member.gradeCode }}</td>
+        <td headers="current cap">{{ member.capacityPercentage | round }}%</td>
+        <td headers="points cap">{{ member.availablePoints }}</td>
+        <td headers="hrs contracted">{{ member.contractedHours }}</td>
+        <td headers="hrs reduction">{{ member.reductionHours }}</td>
+        <td headers="cases total">{{ member.totalCases }}</td>
+      </tr>
+        {% endfor%}
```
andrew-js-wright: indentation of closing `endfor` should be consistent with opening `for`.

app/views/overview.html
```diff
@@ -12,38 +12,9 @@
...
-        <th scope="row">Reduction</th>
-        <td class="numeric">{{ overviewDetails.reduction }}</td>
-        </tr>
-    </tbody>
-    </table>
-  </div>
-</div>
+{% include "includes/individual-overview.html" %}
+
+{% include "includes/team-overview-table.html" %}
```
andrew-js-wright: I see the team view being extended so it can display any table which is a higher level than an individual. It might be worth having an if/case statement here so it's obvious that only one partial is actually being loaded.

app/views/workload-points.html
```diff
@@ -37,7 +37,7 @@ <h2 class="heading-large">
             value="{{ workloadPoints['nominalTargetPointsSPO'] }}">
   </div>
 
-  {% include "partials/csrf-hidden-input.html" %}
+  {% include "includes/csrf-hidden-input.html" %}
```
andrew-js-wright: 👍

test/accessibility/run-pa11y.sh
```diff
@@ -46,6 +46,7 @@ urls+=("$host/region/$regionId/$case_progress_url")
 # Overview
 overview_url='overview'
 urls+=("$host/offender-manager/$omId/$overview_url")
+urls+=("$host/team/$teamId/$overview_url")
```
andrew-js-wright: 👍

test/e2e/view-team-case-progress.js
```diff
@@ -0,0 +1,26 @@
+const expect = require('chai').expect
+
+const workloadCapacityHelper = require('../helpers/data/aggregated-data-helper')
+
+var teamId
+
+describe('View Team overview', function () {
```
andrew-js-wright: Capital `O` on overview +1 pedantic points for me :)

test/integration/services/data/test-get-team-caseload-overview.js
```diff
@@ -0,0 +1,39 @@
+const expect = require('chai').expect
+
+const workloadCapacityHelper = require('../../../helpers/data/aggregated-data-helper')
+const getTeamCaseloadOverview = require('../../../../app/services/data/get-team-caseload-overview')
+
+var inserts = []
+
+describe('services/data/get-team-caseload-overview', function () {
+  before(function (done) {
```
andrew-js-wright: After our discussion yesterday, can we move away from calling `done()` and instead return the promise so that promises don't cause a timeout. This applies to all other tests.

test/unit/services/test-get-overview.js
```diff
@@ -56,4 +64,18 @@ describe('services/get-overview', function () {
       done()
     })
   })
+
+  it('should return a results object with the correct overview details for a team', function (done) {
+    var teamName = 'team'
```
andrew-js-wright: I know this is copied from the existing test but can we get these values from `constants/organisation-unit.js` so we're not repeating ourselves.

app/assets/sass/_wmt/data-table.scss
```diff
@@ -0,0 +1,162 @@
...
+}*/
+
+table.settings td:first-child{
+    // width:125px;
+}
+
+table.settings td:nth-child(2) input{
+    // width:100px;
+}
+
```
dconey646: Yea, accidental copy and paste from the prototype

app/assets/sass/_wmt/data-table.scss
```diff
@@ -0,0 +1,162 @@
...
+
+
+
+ul.sub-tier-header {
+width: 369px;
+display: block;
+float: left;
+margin: -5px -10px -5px -10px;
```
dconey646: Same as above really..

app/assets/sass/application.scss
```diff
@@ -35,3 +35,4 @@ $path: "/public/images/";
 
 @import '_wmt/wmt-page-heading';
 @import '_wmt/wmt-sub-nav';
+@import '_wmt/data-table';
```
dconey646: Yep, good spot.

app/services/data/get-team-caseload-overview.js
dconey646: Yea, good spot.

app/views/overview.html
dconey646: Yea, so part of the next story, we are going to refactor that to org unit overview (or something like that).

We tried the if/case statement, but the way in which nunjucks renders the templates, it appears to render the includes before it performs the if statements. There is an if statement  at the top of the include files to cover this.

test/e2e/view-team-case-progress.js
dconey646: Pedantic points are the best points :)

test/unit/services/test-get-overview.js
dconey646: Yea, thats a good idea going forward, will do that for all tests where we call this.

test/integration/services/data/test-get-team-caseload-overview.js
dconey646: Good call. I will apply this to other tests now as well.

app/services/data/get-individual-overview.js
```diff
@@ -14,7 +14,7 @@ module.exports = function (id) {
            'workload_points_calculations.available_points AS availablePoints',
            'workload_points_calculations.total_points AS totalPoints',
            'workload.total_cases AS cases',
-           'workload_points_calculations.contracted_hours AS contractedHours',
+           'workload_owner.contracted_hours AS contractedHours',
```
tsheppard01: Not sure what your standard approach for this is , but could do this in the view using a COALESCE.  Then application doesn't need to worry, but means pushing logic into the DB, which may not be ideal.

app/views/overview.html
tsheppard01: Would it be worth combining the individual-overview and team-overview-table into a single include?   Would mean all logic regarding which table is displayed is in a single place rather than spread out over 2.

test/helpers/data/aggregated-data-helper.js
```diff
@@ -230,6 +230,15 @@ module.exports.selectIdsForWorkloadOwner = function () {
   return promise
 }
 
+module.exports.getTeamId = function () {
```
tsheppard01: To me getTeamId implies you are getting the Id of a specific team.  Perhaps rename this.  Maybe to getAnyExistingTeamId? Or something else sensible.

app/views/overview.html
dconey646: ^I like @tsheppard01's idea. I think that would be a good implementation.

test/helpers/data/aggregated-data-helper.js
dconey646: Yep, getAnyExistingTeamId sounds good.

app/services/data/get-individual-overview.js
dconey646: Going to create and extra method in the probation rules that will get the default contracted hours if they are null.

tsheppard01:
LGTM 👍

PR #74, opened 2017-08-09
Feature/wmt 784/merge

caoimheb:


app/services/data/get-team-caseload.js
```diff
@@ -0,0 +1,27 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+module.exports = function (id) {
+  var table = 'team_caseload_overview'
+  var whereObject = {}
+  if (id !== undefined) {
```
caoimheb: Should this be the case for team? This was included in the other queries to account for national level not having an id, but I don't think we ever want to query the team table with no filter as we will get incorrect results.

test/integration/services/data/test-get-team-caseload.js
```diff
@@ -0,0 +1,36 @@
...
+    return getTeamCaseload(inserts.filter((item) => item.table === 'team')[1].id, 'team')
+      .then(function (results) {
+        expect(results.length).to.eql(1)
+        expect(results[0].gradeCode).to.eql('PO')
+        expect(results[0].untiered).to.eql(30)
+        expect(results[0].d2).to.eql(30)
+        expect(results[0].d1).to.eql(30)
+        expect(results[0].c2).to.eql(30)
+        expect(results[0].c1).to.eql(30)
+        expect(results[0].b2).to.eql(30)
```
caoimheb: Worth changing test data to different values (not all 30) so we make sure fields are being assigned correctly?

app/services/data/get-team-caseload.js
```diff
@@ -0,0 +1,27 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+module.exports = function (id) {
+  var table = 'team_caseload_overview'
```
caoimheb: Think this view should be called `team_caseload_view`. This would also require an update to the view name in the seeds folder in `wmt-worker`

app/services/data/get-team-caseload.js
ddebella: As you said since we don't currently have a plan for the Regional level we want to ensure the data is filtered by id.

test/integration/services/data/test-get-team-caseload.js
ddebella: Agreed, and this is now addressed.

app/services/data/get-team-caseload.js
ddebella: The the view name is renamed as well as the actual filename.

PR #80, opened 2017-08-14
Ignore pa11y HTML report outputs

tswann:
Add .gitignore file for test/accessibility outputs


test/accessibility/.gitignore
```diff
@@ -0,0 +1 @@
+*.html
```
caoimheb: Pa11y no longer outputs html reports, which is why this was removed. Guessing you had some old ones in your local source? So worth leaving this in?

tswann: 👍

PR #82, opened 2017-08-14
787: Adding probation rules and calculation contracted hours for the …

dconey646:
…workload owner. Both individual and team. Updated tests

### DIFF
--- a/app/services/data/get-individual-overview.js
+++ b/app/services/data/get-individual-overview.js
@@ -6,6 +6,7 @@ module.exports = function (id) {
     .join('team', 'workload_owner.team_id', 'team.id')
     .join('workload', 'workload.workload_owner_id', 'workload_owner.id')
     .join('workload_points_calculations', 'workload_points_calculations.workload_id', 'workload.id')
+    .join('workload_points', 'workload_points_calculations.workload_points_id', 'workload_points.id')
     .join('workload_report', 'workload_points_calculations.workload_report_id', 'workload_report.id')
     .join('offender_manager', 'offender_manager.id', 'workload_owner.offender_manager_id')
     .first('offender_manager.grade_code AS grade',
@@ -15,7 +16,9 @@ module.exports = function (id) {
            'workload_points_calculations.total_points AS totalPoints',
            'workload.total_cases AS cases',
            'workload_owner.contracted_hours AS contractedHours',
-           'workload_points_calculations.reduction_hours AS reduction')
+           'workload_points_calculations.reduction_hours AS reduction',
+           'workload_points.default_contracted_hours_po AS defaultContractedHoursPo',
+           'workload_points.default_contracted_hours_pso AS defaultContractedHoursPso')
     .where('workload_owner.id', id)
     .whereNot('workload_report.effective_from', null)
     .where('workload_report.effective_to', null)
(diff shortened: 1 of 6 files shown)


app/services/get-overview.js
```diff
@@ -16,18 +18,34 @@ module.exports = function (id, organisationLevel) {
...
-    }
+    result.overviewDetails = calculateCapacityPercentage(results)
     result.breadcrumbs = getBreadcrumbs(id, organisationLevel)
     result.title = result.breadcrumbs[0].title
     result.subTitle = organisationalUnitType.displayText
     return result
   })
 }
+
+var calculateCapacityPercentage = function (results) {
```
caoimheb: Don't think `calculateCapacityPercentage` should be also calling `getContractedHours`. Possibly a rename? Or take contracted hours out of this function

app/services/data/get-individual-overview.js
```diff
@@ -15,7 +16,9 @@ module.exports = function (id) {
            'workload_points_calculations.total_points AS totalPoints',
            'workload.total_cases AS cases',
            'workload_owner.contracted_hours AS contractedHours',
-           'workload_points_calculations.reduction_hours AS reduction')
+           'workload_points_calculations.reduction_hours AS reduction',
+           'workload_points.default_contracted_hours_po AS defaultContractedHoursPo',
+           'workload_points.default_contracted_hours_pso AS defaultContractedHoursPso')
```
caoimheb: Still not completely sure on best way to proceed with these bands. Currently they only have one default hours for all bands. Do we confine it to one? If we want more, do we need to include all? Just a thought...

test/unit/services/test-get-overview.js
```diff
@@ -16,7 +16,10 @@ const OVERVIEW = {
   totalPoints: 40,
   cases: '2',
   hours: '3',
-  reduction: '4'
+  reduction: '4',
+  contractedHours: 0,
```
caoimheb: Should `contractedHours` returned in the results object from `getOverview` not have been replaced with the default value?

package.json
```diff
@@ -85,6 +85,7 @@
     "serve-favicon": "2.4.x",
     "underscore": "^1.8.3",
     "util": "^0.10.3",
-    "validator": "^7.0.0"
+    "validator": "^7.0.0",
+    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#tech/wmt-787/make-default-contracted-hours-more-generic"
```
dconey646: Yea, so once probation rules is merged in, I will change to the newest version before this PR gets merged.

app/services/get-overview.js
dconey646: Good shout. I will rename

test/unit/services/test-get-overview.js
dconey646: Yep, it makes more sense to have default hours.

app/services/get-overview.js
```diff
@@ -16,18 +18,38 @@ module.exports = function (id, organisationLevel) {
...
-    }
+    result.overviewDetails = calcualteValues(results)
     result.breadcrumbs = getBreadcrumbs(id, organisationLevel)
     result.title = result.breadcrumbs[0].title
     result.subTitle = organisationalUnitType.displayText
     return result
   })
 }
+
+var calcualteValues = function (results) {
```
caoimheb: Typo in function name

PR #85, opened 2017-08-15
Feature/wmt 820/get caseload by type

ddebella:
Ready for review

### DIFF
--- a/test/unit/services/test-get-caseload.js
+++ b/test/unit/services/test-get-caseload.js
@@ -8,9 +8,15 @@ const orgUnitFinder = require('../../../app/services/helpers/org-unit-finder')
 
 const breadcrumbHelper = require('../../helpers/breadcrumb-helper')
 
+var CASE_TYPE = {
+  COMMUNITY: 'COMMUNITY',
+  CUSTODY: 'CUSTODY',
+  LICENSE: 'LICENSE'
+}
+
 const CASELOAD = {
   name: 'Todd Umptious',
-  caseType: 'COMMUNITY',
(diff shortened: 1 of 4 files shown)


test/unit/services/test-get-caseload.js
```diff
@@ -24,7 +25,12 @@ const CASELOAD = {
   a: 0
 }
 
-const TEAM_CASELOAD = [CASELOAD, CASELOAD]
+const OVERALL_CASELOAD = Object.assign({}, CASELOAD, {totalCases: 9, untiered: 3, d2: 3, c1: 3, b1: 3})
+const COMMUNITY_CASELOAD = Object.assign({}, CASELOAD, {caseType: 'COMMUNITY'})
+const CUSTODY_CASELOAD = Object.assign({}, CASELOAD, {caseType: 'CUSTODY'})
+const LICENSE_CASELOAD = Object.assign({}, CASELOAD, {caseType: 'LICENSE'})
```
ddebella: We already have the default values in the CASELOAD object. The overall alters the default values by summing those default values.

test/unit/services/test-get-caseload.js
```diff
@@ -24,7 +25,12 @@ const CASELOAD = {
   a: 0
 }
```
caoimheb: Could you include multiple workload owners' caseloads of each type? Just to make sure queries / functions work for larger data sets?

ddebella: I have added a two more data set to the test case and updated the overall check

PR #86, opened 2017-08-16
Feature/wmt 820/case by type screens

caoimheb:


app/views/caseload.html
```diff
@@ -12,47 +12,78 @@
...
+
+{% block body_end %}
+  <script src="/public/javascripts/vendor/jquery.js"></script>
+  <script src="/public/javascripts/jquery.tabs.js"></script>
+
+  <script>
+    $(function(){
+      $('.js-tabs').tabs();
+    });
+  </script>
```
dconey646: Really like this implementation, it looks cool 👍

PR #89, opened 2017-08-17
822: Update data service for LDU caseload

caoimheb:
Replace get-team-caseload with a more generic get-caseload. Add caseloadView
to the org unit constant to allow query to call the correct view.

Add integration tests for LDU level caseload. Update data helper to insert
offender managers of both PO and PSO, for more realistic tests


app/services/data/get-caseload.js
```diff
@@ -0,0 +1,35 @@
...
+  if (orgUnit.name === orgUnitConstants.TEAM.name) {
+    selectColumns.push('case_type AS caseType')
+  }
+
+  return knex(table)
+    .where('id', id)
+    .select(selectColumns)
+    .then(function (results) {
+      return results
+    })
```
dconey646: 👍

package.json
```diff
@@ -7,7 +7,7 @@
     "start": "gulp",
     "build": "yarn install && gulp generate-assets && gulp sync",
     "unit-test": "standard && mocha --recursive test/unit/",
-    "integration-test": "gulp assets && mocha --recursive test/integration --timeout=10000",
+    "integration-test": "gulp assets && mocha --recursive test/integration --timeout=30000",
```
dconey646: How come an increase in the timeout?

test/helpers/data/aggregated-data-helper.js
```diff
@@ -119,93 +141,116 @@ var addLdu = function (inserts) {
...
-      })
-      return inserts
+    return knex('tiers').returning('id').insert(tiers)
+  })
+  .then(function (ids) {
+    ids.forEach((id) => {
+      inserts.push({table: 'tiers', id: id})
     })
+    return inserts
+  })
```
dconey646: 👍

package.json
caoimheb: The data inserts and removes were timing out on me. Not sure if it's maybe because I'm on a worse network when WFH or something.

PR #90, opened 2017-08-17
Feature/wmt 821/caseload summary by type

ddebella:
Ready for review.

### DIFF
--- a/test/e2e/view-caseload.js
+++ b/test/e2e/view-caseload.js
@@ -29,6 +29,22 @@ describe('View your caseload flow', () => {
         })
     })
 
+    it('with the correct caseload total summary for each case type', function () {
+      return browser.url(teamCaseloadUrl)
+        .getText('#custodyTotal')
+        .then(function (text) {
+          expect(text).to.eql('315\nCustody cases')
+        })
+        .getText('#communityTotal')
+        .then(function (text) {
+          expect(text).to.eql('315\nCommunity cases')
+        })
+        .getText('#licenseTotal')
+        .then(function (text) {
+          expect(text).to.eql('315\nLicense cases')
+        })
+    })
+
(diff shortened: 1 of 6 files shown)


test/e2e/view-caseload.js
```diff
@@ -29,6 +29,22 @@ describe('View your caseload flow', () => {
         })
     })
 
+    it('with the correct caseload total summary for each case type', function () {
+      return browser.url(teamCaseloadUrl)
+        .getText('#custodyTotal')
+        .then(function (text) {
+          expect(text).to.eql('315\nCustody cases')
```
caoimheb: This number is dependent on seed data, so this test will fail every time we update the seed files? Maybe better to just confirm that a total for each is displayed, and the unit tests will verify it will be the correct total

test/unit/services/test-get-caseload.js
```diff
@@ -117,4 +117,31 @@ describe('services/get-caseload', function () {
       expect(result.licenseCaseloadDetails[0]).to.eql(LICENSE_CASELOAD)
     })
   })
+
+  it('should call get-team-caseload and return caseload total summary for CUSTODY casetype for a team', function () {
+    var teamName = orgUnitConstant.TEAM.name
+    getTeamCaseload.withArgs(id).resolves(TEAM_CASELOAD)
+    return getCaseload(id, teamName).then(function (result) {
+      assert(getTeamCaseload.called)
+      expect(result.custodyTotalSummary).to.eql(9)
```
caoimheb: Update data so values for each total are different, to show explicitly that the correct numbers are being added?

test/e2e/view-caseload.js
dconey646: These could vary could they not?

PR #104, opened 2017-08-25
823: Add reduction page

ddebella:


app/views/add-reduction.html
```diff
@@ -0,0 +1,168 @@
...
+
+<!-- TEMPORARY VALUE START -->
+{% set isViewOnly=false %}
+
+<!-- TEMPORARY VALUE END -->
+
+<main id="content" role="main">
+  {% include "includes/breadcrumbs.html" %}
+  
+  {% if isViewOnly %}
```
caoimheb: Maybe a different name for this var, like `isAdd`? Since add is the different case, whereas the screen will then look the same for view, edit, archive, delete

app/views/add-reduction.html
```diff
@@ -0,0 +1,168 @@
...
+  {% else %} 
+    {% set reductionTitle='New reduction' %} 
+    {% set reductionDescription='Add a new reduction for' %}
+    {% set method='POST' %}
+  {% endif %}
+
+  <header class="heading-xlarge" style="margin-top:0; margin-bottom:0">
+    {{ reductionTitle }}
+  </header>
+  <p class="lede">{{ reductionDescription }} <span class="bold">{{ title }}</span></p>
```
caoimheb: 👍

app/views/add-reduction.html
```diff
@@ -0,0 +1,168 @@
...
+        {% for reason in referenceData %}
+        <option value="{{ reason.reasonShortName }}">{{ reason.reasonShortName }}</option>)
+        {% endfor %}
+      </select>
+    </div>
+
+    <div class="form-group" id="reduction_hours">
+      <label class="form-label-bold" for="hours">Hours</label>
+      <input class="form-control" id="hours" name="hours" type="text" style="width: 100px"      
+      {% if isViewOnly %} value="{{ result.view.hours }}" {% endif %}>
```
caoimheb: These aren't needed for the add story?

PR #109, opened 2017-09-01
View all reduction types and allow each to be viewed

ddebella:


app/services/data/get-reductions.js
```diff
@@ -0,0 +1,25 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+
+module.exports = function (id) {
+  return knex('reductions')
+    .join('reduction_reason', 'reductions.reduction_reason_id', 'reduction_reason.id')
+    .join('reduction_category', 'reduction_reason.category_id', 'reduction_category.id')
+    .modify(function (whereBlock) {
```
tsheppard01: Could move this where block to above the query to make it a bit clearer?  Like we have in the overview queries.

app/services/helpers/reduction-helper.js
```diff
@@ -0,0 +1,32 @@
...
+  var scheduled = []
+  var archived = []
+  var currentDate = new Date()
+  reductions.filter(function (reduction) {
+    if (currentDate.getTime() > reduction.reductionStartDate.getTime() &&
+            currentDate.getTime() < reduction.reductionEndDate.getTime()) {
+      active.push(reduction)
+    } else if (currentDate.getTime() < reduction.reductionStartDate.getTime()) {
+      scheduled.push(reduction)
+    } else if (currentDate.getTime() > reduction.reductionEndDate.getTime()) {
```
tsheppard01: Are we to have an archived flag in the reductions table.  ie active/scheduled deductions can be archived while the are still active/scheduled?

app/services/helpers/reduction-helper.js
```diff
@@ -0,0 +1,32 @@
+
+var dateFormatter = require('../date-formatter')
+
+module.exports.determineReductionsStatus = function (reductions) {
```
tsheppard01: Minor: Maybe rename to getReductionsByStatus as that is what is returned?

app/services/data/get-reductions.js
```diff
@@ -0,0 +1,25 @@
+const config = require('../../../knexfile').web
+const knex = require('knex')(config)
+
+module.exports = function (id) {
```
tsheppard01: Could you call this variable workloadOwnerId?  Means we don't get confused about whether it is the id of the actual reduction

test/e2e/view-reduction.js
```diff
@@ -0,0 +1,100 @@
...
+        .then(function (text) {
+          expect(text).to.equal('Hours')
+        })
+        .getText('#archived_start_date')
+        .then(function (text) {
+          expect(text).to.equal('Start date')
+        })
+        .getText('#archived_end_date')
+        .then(function (text) {
+          expect(text).to.equal('End date')
```
tsheppard01: 👍

test/unit/services/test-reduction-service.js
```diff
@@ -58,10 +58,12 @@ before(function () {
 describe('services/reductions-service', function () {
```
tsheppard01: Have you added integration tests for the getReduction stuff?

app/views/reductions.html
```diff
@@ -14,88 +14,102 @@
...
+                <th id="active_type">{{ columnNameType }}</th>
+                <th id="active_hours">{{ columnNameHours }}</th>
+                <th id="active_start_date">{{ columnNameStartDate }}</th>
+                <th id="active_end_date">{{ columnNameEndDate }}</th>
                 <th></th>
             </tr>
         </thead>
 
         <tbody class="">
             {% if newReduction %}
```
tsheppard01: Where is this newReduction set?

app/services/data/get-reductions.js
```diff
@@ -0,0 +1,25 @@
...
+            'workload_owner_id AS workloadOwnerId',
+            'hours',
+            'reduction_reason_id AS reductionReasonId',
+            'reduction_category.id AS categoryId',
+            'reduction_reason.reason_short_name AS reasonShortName',
+            'reduction_reason.allowance_percentage AS allowancePercentage',
+            'effective_from AS reductionStartDate',
+            'effective_to AS reductionEndDate',
+            'notes',
+            'status'
```
tsheppard01: 👍

app/views/reductions.html
ddebella: It seem when we do a new reduction the call is redirected to this reduction page with a full query returned including the new added reduction. This might not be required as a result.

app/services/data/get-reductions.js
ddebella: Done.

app/services/data/get-reductions.js
ddebella: Done.

app/services/helpers/reduction-helper.js
ddebella: We are to have status flags for Archived and Deleted records. If a record is in either of those two stages then it should not be in Active or Scheduled stage. I will update the query to reflect the status flag.

app/services/helpers/reduction-helper.js
ddebella: Done

PR #111, opened 2017-09-04
Feature/wmt 876/edit reduction

tsheppard01:


app/services/data/get-reduction-by-id.js
```diff
@@ -0,0 +1,15 @@
...
+module.exports = function (id) {
+  var whereObject = {}
+  if (id !== undefined) {
+    whereObject.id = id
+  }
+
+  return knex('reductions')
+    .where(whereObject)
+    .then(function (reduction) {
+      return reduction[0]
```
ddebella: Why is this returning the first element?

app/services/data/update-reduction.js
```diff
@@ -0,0 +1,16 @@
...
+      reduction_reason_id: newReduction.reasonForReductionId,
+      workload_owner_id: workloadOwnerId,
+      hours: newReduction.hours,
+      effective_from: newReduction.reductionStartDate,
+      effective_to: newReduction.reductionEndDate,
+      notes: newReduction.notes
+    })
+    .where('id', reductionId)
+    .returning('id')
+}
```
ddebella: 👍

app/services/reductions-service.js
```diff
@@ -35,12 +37,25 @@ module.exports.getAddReductionsRefData = function (id, organisationLevel) {
...
 
-module.exports.addReduction = function (id, newReduction) {
-  var addReductionPromise = addReduction(id, newReduction)
+module.exports.upsertReduction = function (id, reductionId, reduction) {
+  var reductionPromise
+  if (reductionId !== undefined) {
+    reductionPromise = updateReduction(reductionId, id, reduction)
+  } else {
+    reductionPromise = addReduction(id, reduction)
+  }
```
ddebella: 👍

app/views/add-reduction.html
```diff
@@ -16,19 +20,20 @@
   <p class="lede">Add a new reduction for <span class="bold">{{ title }}</span></p>
```
ddebella: You need to include different title when showing reduction vs adding a reduction. Similar to the if else for heading.

app/views/add-reduction.html
```diff
@@ -6,7 +6,11 @@
   {% include "includes/breadcrumbs.html" %} 
   
   <header class="heading-xlarge" style="margin-top:0; margin-bottom:0">
-    New reduction
+    {% if reduction.reasonId !== undefined %}
```
ddebella: Why do you rely on the reason.id as opposed to the reduction id?

app/views/add-reduction.html
```diff
@@ -16,19 +20,20 @@
...
   <form action="/offender-manager/{{ linkId }}/add-reduction" method=POST id="reductionForm">
+    <input type="hidden" name="reductionId" value="{{reduction.id}}"/>
     <div class="form-group" id="reduction_reason_selection">
       <label class="form-label-bold" for="select-box">Reason for reduction</label>
-      <select class="form-control" id="select-box" name="reasonForReductionId" onChange="populateHours(), calculateExpiry()">        
+      <select class="form-control" id="select-box" name="reasonForReductionId" selected="{{ reduction.reasonId }}" onChange="populateHours(), calculateExpiry()">        
         <option value="" disabled selected>Please select a reason...</option>
         {% for reason in referenceData %}
-        <option value="{{ reason.id }}">{{ reason.reasonShortName }}</option>
+        <option value="{{ reason.id }}" {% if reason.id === reduction.reasonId %} selected {% endif %}>{{ reason.reasonShortName }}</option>
```
ddebella: 👍

app/routes/reductions.js
```diff
@@ -40,22 +41,31 @@ module.exports = function (router) {
...
-        subTitle: result.subTitle,
-        subNav: getSubNav(id, organisationLevel, req.path),
-        referenceData: result.referenceData,
-        failureText: failureText
+    return getAddReductionsReferenceDataPromise
+    .then(function (result) {
+      return reductionsService.getReductionByReductionId(reductionId)
+      .then(function (reduction) {
+        if (reduction !== undefined && reduction.workload_owner_id !== id) {
+          reduction = undefined
```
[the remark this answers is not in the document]
tsheppard01: You could potentially hack the URL to something like, `/offender-manager/123/add-reduction?reductionId=12`.  But if reduction 12 does not belong to offender-manager 123 then we shouldn't show it.  Can you think of a clearer way to do this?

app/routes/reductions.js
```diff
@@ -40,22 +41,31 @@ module.exports = function (router) {
...
       })
     })
   })
 
   router.post('/:organisationLevel/:id/add-reduction', function (req, res, next) {
     var organisationLevel = req.params.organisationLevel
     var id = req.params.id
+    var reductionId = req.body.reductionId
 
     if (organisationLevel !== organisationUnitConstants.OFFENDER_MANAGER.name) {
```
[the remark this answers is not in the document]
tsheppard01: good idea, moved

app/services/data/get-reduction-by-id.js
tsheppard01: I want to return an object rather than an array of objects here

tsheppard01: Update to include fields in the select statement (I realised having no select statement is doing a `select *` and I hate `select *`)

app/routes/reductions.js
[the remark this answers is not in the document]
tsheppard01: Looking at this unearthed a very subtle error, id is being taken as a string rather than an int as I thought so comparison always failed.

app/views/add-reduction.html
tsheppard01: No reason, changed to just look at the reduction, the entire thing is undefined if there is no existing reduction

tsheppard01: Moved all variables that are different between add and edit reduction to a single if at the top of the file

app/views/add-reduction.html
tsheppard01: Add all variable to a single if at top of file

app/services/data/get-reduction-by-id.js
ddebella: select * on a table is not too bad. using * on a command prompt is dangerous (rm *) :-)

PR #112, opened 2017-09-05
Feature/wmt 876/merge

ddebella:
Ready for review.

### DIFF
--- a/app/assets/javascripts/add-reduction.js
+++ b/app/assets/javascripts/add-reduction.js
@@ -10,59 +10,69 @@
 
     var chosenReductionIndex = document.getElementById("select-box").selectedIndex
 
-    // Dummy option in dropdown means array is offset by one.
-    chosenReduction = refData[chosenReductionIndex - 1]
+    if(chosenReductionIndex){
+      // Dummy option in dropdown means array is offset by one.
+      chosenReduction = refData[chosenReductionIndex - 1]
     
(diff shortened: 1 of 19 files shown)


app/routes/reductions.js
```diff
@@ -13,25 +13,28 @@ module.exports = function (router) {
     if (organisationLevel !== organisationUnitConstants.OFFENDER_MANAGER.name) {
```
dconey646: Would be good to have an explicit route for edit reduction. As editing a reduction is essentially viewing. So, could we add that please?

Good for the customer to see correct urls and also for new devs coming on.

app/assets/javascripts/add-reduction.js
```diff
@@ -10,59 +10,69 @@
...
+        document.getElementById('end-day').value = ''
+        document.getElementById('end-month').value = ''
+        document.getElementById('end-year').value = ''
+      }
     }
+  }
+
+  window.onload = function () {
+    populateHours()
+    calculateExpiry()
```
dconey646: How come we are doing this on load? Surely the data we get back for the reduction should the source of truth?

tsheppard01: I was doing this more to set the hours and end date fields to read only,  should I change this to do only that?  Leave the actual values as they are in the returned reduction?

PR #123, opened 2017-09-11
Update web to account for removed grade_code in offender_manager table

caoimheb:
Also fixed failing reductions test


app/services/reductions-service.js
```diff
@@ -15,11 +15,11 @@ module.exports.getReductions = function (id, organisationLevel) {
   var result = {}
   var organisationalUnitType = getOrganisationUnit('name', organisationLevel)
 
-  var getReductionsDataPromise = getReductions(id)
-
   result.breadcrumbs = getBreadcrumbs(id, organisationLevel)
   result.title = result.breadcrumbs[0].title
   result.subTitle = organisationalUnitType.displayText
+
+  var getReductionsDataPromise = getReductions(id)
```
tsheppard01: What was the problem here, why did this need moved?  You could probably remove this Promise variable and start the return below on getReductions(id).  Bit cleaner In my opinion

test/helpers/data/aggregated-data-helper.js
```diff
@@ -97,10 +97,17 @@ module.exports.addWorkloadCapacitiesForOffenderManager = function () {
...
+      var offenderManagerTypes = [
+        { grade_code: 'PO' },
+        { grade_code: 'PSO' }
+      ]
+      return knex('offender_manager_type').returning('id').insert(offenderManagerTypes)
     })
     .then(function (ids) {
-      inserts.push({ table: 'offender_manager_type', id: ids[0] })
+      ids.forEach((id) => {
+        inserts.push({table: 'offender_manager_type', id: id})
```
ddebella: addRegion only inserts one record from the given array. So the inserts array will be all inserted. Does it make sense to keep it as it was?

test/accessibility/run-pa11y.sh
```diff
@@ -26,45 +26,51 @@ fi
...
 # Capacity
 capacity_url='caseload-capacity'
-urls+=("$host/offender-manager/$omId/$capacity_url")
-urls+=("$host/team/$teamId/$capacity_url")
-urls+=("$host/ldu/$lduId/$capacity_url")
-urls+=("$host/region/$regionId/$capacity_url")
+urls+=("$host/$offender_manager/$omId/$capacity_url")
+urls+=("$host/$team/$teamId/$capacity_url")
+urls+=("$host/$ldu/$lduId/$capacity_url")
+urls+=("$host/$region/$regionId/$capacity_url")
```
tsheppard01: 👍

test/unit/services/test-reduction-service.js
```diff
@@ -45,6 +47,33 @@ var referenceData = [
...
+var activeReductions = [ baseReduction ]
+
+var archivedReductions = [
+  Object.assign({}, baseReduction, { id: 2, status: 'ARCHIVED' }),
+  Object.assign({}, baseReduction, { id: 3, status: 'ARCHIVED' })
+]
+
+var reductionsByStatus = {
+  activeReductions: activeReductions,
+  scheduledReductions: [],
```
ddebella: Why not include scheduled reductions too?

test/unit/services/test-reduction-service.js
```diff
@@ -71,15 +106,20 @@ beforeEach(function () {
...
+      return reductionService.getReductions(1, orgUnitConstant.OFFENDER_MANAGER.name)
         .then(function (result) {
+          assert(getReductions.called)
+          assert(getBreadcrumbsStub.called)
           expect(result.title).to.equal('John Doe')
           expect(result.subTitle).to.equal('Offender Manager')
-          assert(getBreadcrumbsStub.called)
+          expect(result.activeReductions).to.eql(activeReductions)
+          expect(result.scheduledReductions).to.eql([])
+          expect(result.archivedReductions).to.eql(archivedReductions)
```
tsheppard01: How are deleted reductions handled?  Are they retireved from database? If so it might be good to test around that.

app/services/data/get-individual-overview.js
```diff
@@ -9,7 +9,8 @@ module.exports = function (id) {
     .join('workload_points', 'workload_points_calculations.workload_points_id', 'workload_points.id')
```
tsheppard01: Should this be in a view.  All other case overview pages are in views and there is alot going on here.  Should more complicated queries be in views?

test/helpers/data/aggregated-data-helper.js
caoimheb: Before it was only adding one offender manager type, which wasn't a problem as we were defining grade_code in the offender_manager rows. Now we are taking grade_code from the offender_manager_type table so need more test data. 

Region adds one, but in the method used for the caseload tests, we specifically add PO and PSO offender managers, so both need to be in the offender_manager_type data

test/unit/services/test-reduction-service.js
caoimheb: The idea was to show arrays of size 0, 1, and greater than 1 being processed

app/services/reductions-service.js
caoimheb: Happy to move. Not completely sure why the ordering needed changed, but it fixed the failing test

test/unit/services/test-reduction-service.js
caoimheb: Good point. The view shouldn't have to deal with them, but the helper might, @ddebella

app/services/data/get-individual-overview.js
caoimheb: Maybe.....

test/unit/services/test-reduction-service.js
ddebella: Deletes are not removed from the table, but are marked with DELETE status. These are checked  by the helper to ensure any DELETE status are not returned.

app/services/data/get-individual-workload-reports.js
```diff
@@ -2,9 +2,7 @@ const config = require('../../../knexfile').web
 const knex = require('knex')(config)
 
 module.exports = function (id, fromDate, toDate) {
-  return knex('workload_points_calculations')
-    .join('workload', 'workload_points_calculations.workload_id', '=', 'workload.id')
-    .join('workload_report', 'workload_points_calculations.workload_report_id', '=', 'workload_report.id')
+  return knex('individual_capacity_view')
     .where('workload_report.effective_from', '>=', fromDate)
     .where('workload_report.effective_from', '<=', toDate)
```
tsheppard01: Should this be done in the view like the others?

app/services/get-capacity-view.js
```diff
@@ -14,11 +12,7 @@ module.exports = function (id, capacityDateRange, organisationLevel) {
     throw new Error(organisationLevel + ' should be offender-manager, region, team, ldu or hmpps')
   }
 
-  if (organisationLevel === routeType.OFFENDER_MANAGER.name) {
-    workloadReportsPromise = getIndividualWorkloadReports(id, capacityDateRange.capacityFromDate.toISOString(), capacityDateRange.capacityToDate.toISOString())
-  } else {
-    workloadReportsPromise = getWorkloadReports(id, capacityDateRange.capacityFromDate.toISOString(), capacityDateRange.capacityToDate.toISOString(), organisationLevel)
-  }
+  workloadReportsPromise = getWorkloadReports(id, capacityDateRange.capacityFromDate.toISOString(), capacityDateRange.capacityToDate.toISOString(), organisationLevel)
```
tsheppard01: Do we not still need to have separate individual and organisation?

app/services/data/get-individual-workload-reports.js
caoimheb: Date range as opposed to getting active workload. Think it needs to be done here

app/services/get-capacity-view.js
caoimheb: I don't think so. The group query should be ok to call the individual view now that it exists. I have updated how the individual query is called (new param, 'offender-manager') and added the view name to the offender-manager constant. This got all tests passing and screens look ok

PR #124, opened 2017-09-12
877: Added unit test for reduction helper

ddebella:


app/services/helpers/reduction-helper.js
```diff
@@ -1,24 +1,30 @@
...
-    }
-    formatReductionDates(reduction)
-  })
+  if (reductions instanceof Array) {
+    reductions.filter(function (reduction) {
+      if ((reduction.reductionStartDate instanceof Date) &&
+        (reduction.reductionEndDate instanceof Date)) {
+        if ((reduction.status !== reductionStatusType.ARCHIVED) && (reduction.status !== reductionStatusType.DELETED) &&
+          (currentDate.getTime() > reduction.reductionStartDate.getTime()) &&
+          (currentDate.getTime() < reduction.reductionEndDate.getTime())) {
```
caoimheb: Can we just use a check for reductionStatusType.SCHEDULED / ACTIVE here instead of the time range check, as the worker should ensure all statuses are correct in the DB?

test/unit/services/helpers/test-reduction-helper.js
```diff
@@ -0,0 +1,31 @@
...
+describe('services/helpers/reduction-helper', function () {
+  describe('getReductionsByStatus', function () {
+    it('should return reductions by status of active archived and scheduled', function () {
+      var result = reductionHelper.getReductionsByStatus(REDUCTIONS)
+      expect(result.activeReductions.length).to.eql(1)
+      expect(result.activeReductions[0]).to.eql(ACTIVE_REDUCTION)
+      expect(result.scheduledReductions.length).to.eql(1)
+      expect(result.scheduledReductions[0]).to.eql(SCHEDULED_REDUCTION)
+      expect(result.archivedReductions.length).to.eql(1)
+      expect(result.archivedReductions[0]).to.eql(ARCHIVE_REDUCTION)
```
caoimheb: 👍

app/services/helpers/reduction-helper.js
ddebella: I would have been more comfortable to remove the date checks if the status attribute was not nullable. However we are not able to set the status to not nullable. As we rely on the reduction worker to set the status of a reduction I am ok relying on that and just check the status and not use the dates in the helper.

PR #129, opened 2017-09-19
Feature/wmt 1028/input validation

tswann:
- Field Validator class to provide isRequired, isInt and isFloat server-side field validation
- Added specific server side validation to Contracted Hours to demonstrate validation changes
- Updated route helper to enable session cookie support in unit tests
- Unit tests updated

### DIFF
--- a/app/routes/contracted-hours.js
+++ b/app/routes/contracted-hours.js
@@ -17,10 +17,7 @@ module.exports = function (router) {
 
     return contractedHoursService.getContractedHours(id, organisationLevel)
     .then(function (result) {
-      var errors = req.session.contractedHourErrors
-      delete req.session.contractedHourErrors
       return res.render('contracted-hours', {
-        errors: errors,
         title: result.title,
         subTitle: result.subTitle,
         breadcrumbs: result.breadcrumbs,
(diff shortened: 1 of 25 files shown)


app/routes/contracted-hours.js
```diff
@@ -36,11 +43,34 @@ module.exports = function (router) {
       return res.sendStatus(404)
     }
 
+    try {
+      if (updatedHours) isValid(updatedHours, next)
```
andrew-js-wright: My understanding here is: if updatedHours is undefined, or otherwise falsey, the validation will not fire and there will be an attempt at writing it to the database.

app/views/contracted-hours.html
```diff
@@ -20,11 +20,15 @@ <h2 class="heading-large">Contracted hours</h2>
 
 <p class="lede">Change the contracted hours for <span class="bold">{{title}}</span></p>
 
+{% if errors['hours'][0] %}
+  <span class="error-message">{{ errors['hours'][0] }}</span>
+{% endif %}
+
 <form action="/offender-manager/{{woId}}/contracted-hours" method="post" class="sln-form-action">
   <input type="hidden" name="sub_page" value="true">
   <div class="form-group">
```
andrew-js-wright: The error is rendered as a link which has a href to the anchor 'hours'. In order for this to work correctly there should be a corresponding element with 'hours' as the id.

PR #131, opened 2017-09-20
Feature/wmt 989/cms reductions

tsheppard01:
Adds capacity breakdown table and stops cms reductions being shown in reductions page and stops cms reductions being in drop down menu in add/edit reductions page


test/integration/services/data/test-get-reduction-reasons.js
```diff
@@ -31,7 +33,12 @@ describe('services/data/get-reduction-reasons', function () {
   it('should return an array of reductions reasons ref data, including category', function () {
```
caoimheb: Add some detail about cms to the test description?

app/services/get-capacity-view.js
```diff
@@ -1,25 +1,85 @@
...
+      if (organisationMap.has(workloadReport.name)) {
+        valueToAdd = organisationMap.get(workloadReport.name)
+        valueToAdd.push(workloadReport)
+      } else {
+        valueToAdd = [workloadReport]
+      }
+      organisationMap.set(workloadReport.name, valueToAdd)
+    })
+
+    organisationMap.forEach(function (reports, orgName) {
```
caoimheb: Is this returned result the key value pair from the map? Should it be the other way around?

app/views/capacity.html
```diff
@@ -49,4 +117,8 @@
     </script>
     <script src="/public/javascripts/plotly.js"></script>
     <script src="/public/javascripts/caseload-capacity.js"></script>
+    <script src="/public/javascripts/round-to-one-decimal-place.js"></script>
```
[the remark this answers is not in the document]
tsheppard01: Thanks

app/routes/capacity-route.js
```diff
@@ -13,6 +14,14 @@ module.exports = function (router) {
       id = req.params.id
     }
 
+    var orgUnit = getOrganisationUnit('name', organisationLevel)
+    var childOrgUnit
+    if (organisationLevel !== organisationUnit.OFFENDER_MANAGER.name) {
+      childOrgUnit = getOrganisationUnit('name', orgUnit.childOrganisationLevel)
```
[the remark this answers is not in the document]
tsheppard01: cool, changed

test/helpers/data/ref-data-helper.js
```diff
@@ -15,19 +15,26 @@ module.exports.addReductionsRefData = function (maxId) {
...
-    inserts.push({table: 'reduction_reason', id: ids[0].id})
-    return inserts
-  })
+    .then(function (ids) {
+      inserts.push({table: 'reduction_reason', id: ids[0].id})
+      return knex('reduction_reason')
+      .select('id')
+      .where('id', (maxId + 2))
+      .then(function (ids) {
+        inserts.push({table: 'reduction_reason', id: ids[0].id})
```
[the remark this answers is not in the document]
tsheppard01: yip, changed

test/integration/services/data/test-get-capacity-breakdown.js
```diff
@@ -0,0 +1,83 @@
+const expect = require('chai').expect
+
+const dataHelper = require('../../../helpers/data/aggregated-data-helper')
+const getCapacityBreakdown = require('../../../../app/services/data/get-capacity-breakdown')
+
+var inserts = []
+
+describe('services/data/get-workload-reports-for-org', function () {
```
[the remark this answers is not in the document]
tsheppard01: copy pasta

test/integration/services/data/test-get-capacity-breakdown.js
```diff
@@ -0,0 +1,83 @@
...
+  })
+
+  it('should retrieve all workload_owners workloads within a team', function () {
+    return getCapacityBreakdown(inserts.filter((item) => item.table === 'team')[0].id, 'team')
+      .then(function (results) {
+        expect(results.length).to.be.eql(2)
+        expect(results[0].name).to.be.eql('Test_Forename Test_Surname')
+        expect(results[0].grade).to.be.eql('PO')
+        expect(results[0].total_points).to.be.a('Number')
+        expect(results[0].reduction_hours).to.be.a('Number')
```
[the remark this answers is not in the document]
tsheppard01: Yip, changed

test/integration/services/data/test-get-reduction-reasons.js
tsheppard01: updated

app/services/get-capacity-view.js
tsheppard01: I also thought this, but it def comes through as value,key

caoimheb: Strange

test/integration/services/data/test-get-capacity-breakdown.js
```diff
@@ -0,0 +1,98 @@
...
+      var expectedEntry = {
+        total_points: 70,
+        available_points: 35,
+        reduction_hours: 6,
+        linkId: inserts.filter((item) => item.table === 'region')[0].id,
+        name: 'Test Region',
+        grade: 'PO',
+        totalCases: 10,
+        cmsReductionHours: 0,
+        contractedHours: 75
```
caoimheb: Three of these expected entries are the same except for `linkId`? Define in one place and use Object.assign for the `linkId`?

app/views/capacity.html
```diff
@@ -40,6 +40,74 @@
 
 <div id="plotly-div-line" role="presentation"></div>
 
+{% if organisationLevel !== 'offender-manager' %}
+  
+  <div id="team-capacity"></div>
```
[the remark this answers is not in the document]
tsheppard01: Removed

test/integration/services/data/test-get-capacity-breakdown.js
tsheppard01: changed

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
 
 <div id="plotly-div-line" role="presentation"></div>
 
+{% if organisationLevel !== 'offender-manager' %}
+  <h3 class="heading-medium">Capacity breakdown</h3>
+  <table id="example" class="data-table dataTable">
+    <thead>
+      <tr>
+        <th colspan="2" id="team">{{ childOrganisationLevelDisplayText }}</th>
```
andrew-js-wright: `id="team"` here I'm guessing will only be true at an LDU level. Perhaps something like `id="child"` would be more appropriate?

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+        <th class="" colspan="14" id="breakdown">Capacity breakdown</th>
```
andrew-js-wright: Any need for an empty `class` attribute?

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+      </tr>
+      <tr>
+        <th id="name">Name</th>
+        <th id="grade">Grade</th>
+        <th id="capacity">Capacity(%)</th>
+        <th id="cases">Cases</th>
+        <th id="paroms"><abbr title="Parole assessment reports by the Offender Manager">PAROMs</abbr></th>
+        <th id="fdrs"><abbr title="Fast delivery reports">FDRs</abbr></th>
+        <th id="sdrs"><abbr title="Standard delivery reports">SDRs</abbr></th>
+        <th id="arms"><abbr title="Active Risk Management System">ARMS</abbr></th>
```
andrew-js-wright: All of the rest of these abbreviation expansions only capitalise the first word. I think capitalising each make sense but consistency is key.

app/views/capacity.html
andrew-js-wright: Should `colspan` here not be 8 as there are only 10 columns and 2 of them are consumed by the preceding header.?

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+        <th id="gs"><abbr title="Group supervision">GS</abbr></th>
+        <th id="cms"><abbr title="Case management support">CMS</abbr></th>
+      </tr>
+    </thead>
+    {% if organisationLevel === 'team' %}
+      <tbody>
+        {% for member in capacityBreakdown %}
+        <tr>
+          <td headers="name team"><a href="/offender-manager/{{ member.linkId }}">{{ member.name }}</a></td>
```
andrew-js-wright: If the above change was implemented `team` -> `child` then these references should also be updated.

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+          <td headers="grade team">{{ member.grade }}</td>
+          <td headers="capacity breakdown" class="sln-to-round">{{ member.capacityPercentage }}</td>
+          <td headers="cases breakdown">{{ member.totalCases }}</td>
+          <td headers="paroms breakdown">-</td>
+          <td headers="fdrs breakdown">-</td>
+          <td headers="sdrs breakdown">-</td>
+          <td headers="arms breakdown">-</td>
+          <td headers="gs breakdown">-</td>
+          <td headers="cms breakdown" class="sln-to-round">{{ member.cmsPercentage }}</td>
```
andrew-js-wright: We've previously done rounding using a filter template - means our non-javascript using friends also get it. Checkout https://mozilla.github.io/nunjucks/templating.html#round and and example in the code: `includes/organisation-overview.html:16`

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+        </tr>
+        {% endfor %}
+      </tbody>
+    </table>
+  {% else %}
+      <tbody>
+      {% for member in capacityBreakdown %}
+        {% for i in range(0, member.grades.length) %}
```
andrew-js-wright: Should this not be `range(0, member.grades.length-1)` as we're starting from 0?

app/views/capacity.html
tsheppard01: Copy pasta error I believe, fixed

app/views/capacity.html
tsheppard01: More copy pasta, removed and ammended

app/views/capacity.html
tsheppard01: Changed to capitalise the abbreviation words

app/views/capacity.html
tsheppard01: updated

app/services/get-capacity-view.js
```diff
@@ -1,25 +1,85 @@
...
-  return workloadReportsPromise.then(function (results) {
+  return getWorkloadReports(id, capacityDateRange.capacityFromDate.toISOString(), capacityDateRange.capacityToDate.toISOString(), organisationLevel)
+  .then(function (results) {
+    result.capacityBreakdown = []
     result.capacityTable = tableCreator.createCapacityTable(id, organisationalUnitType, capacityDateRange, results)
     result.title = result.breadcrumbs[0].title
     result.subTitle = organisationalUnitType.displayText
+
+    if (organisationalUnitType !== organisationConstant.OFFENDER_MANAGER) {
+      return getCapacityBreakdown(id, organisationLevel)
```
andrew-js-wright: The `getCapacity` function takes into account the start and end dates selected. I'm assuming the breakdown should also do this?

app/services/helpers/percentage-calculator.js
```diff
@@ -1,37 +1,4 @@
...
-      percentageGradeRecords.push(newGradeRecord)
-    }
-    newTeamEntry = Object.assign({}, newTeamEntry, {grades: percentageGradeRecords})
-    percentageResults.push(newTeamEntry)
-  }
-  return percentageResults
-}
-
-var calculatePercentage = function (value, total) {
+module.exports.calculatePercentage = function (value, total) {
```
andrew-js-wright: It doesn't feel that this warrants it's own file any more. Writing `percentageCalculator.calcuatePercentage(n, m)` seems like more hassle that `n / m * 100`.

app/views/capacity.html
tsheppard01: Cool, removed the javascript and removed the class from here and from `caseload.html`

app/views/capacity.html
tsheppard01: `range(min,max)` generates values min, min+1,...,max-1 I believe

andrew-js-wright: Good to know, thanks :)

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
...
+          <tr>
+            {% if i === 0 %}
+              <td rowspan="{{ member.grades.length }}" headers="name team"><a href="/{{ childOrganisationLevel }}/{{ member.linkId }}">{{ member.name }}</a></td>
+            {% endif %}
+          <td headers="grade team">{{ member.grades[i].grade }}</td>
```
andrew-js-wright: These references to team should also be updated to `org`

app/services/get-capacity-view.js
tsheppard01: The capacity breakdown shows the capacity in the current workload, while the getCapcity get the history for the graph.  The capacity breakdown expands on the current capacity column in the overview

app/services/helpers/percentage-calculator.js
tsheppard01: There is some error checking in there to prevent divide by zero, don't want to repeat that every time.

app/services/get-capacity-view.js
andrew-js-wright: That makes sense. Thanks for clarifying.

app/services/helpers/percentage-calculator.js
andrew-js-wright: Good shout, cheers for justifying.

app/views/capacity.html
tsheppard01: Good spot, changed

app/services/data/get-capacity-breakdown.js
```diff
@@ -0,0 +1,29 @@
...
+
+module.exports = function (id, type) {
+  var orgUnit = orgUnitFinder('name', type)
+
+  var table = orgUnit.capacityBreakdownView
+  var whereObject = {}
+
+  if (id !== undefined) {
+    whereObject.id = id
+  }
```
andrew-js-wright: I don't think the `whereObject` is required here as the `id` will come from the path param and therefore always be present.

app/services/data/get-capacity-breakdown.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return knex(table)
+    .where(whereObject)
+    .select('total_points',
+            'available_points',
+            'reduction_hours',
```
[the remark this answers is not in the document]
tsheppard01: Yip, changed

app/views/capacity.html
```diff
@@ -40,6 +40,71 @@
 
 <div id="plotly-div-line" role="presentation"></div>
 
+{% if organisationLevel !== 'offender-manager' %}
+  <h3 class="heading-medium">Capacity breakdown</h3>
+  <table id="example" class="data-table dataTable">
```
andrew-js-wright: `id=example` looks like a copy and paste typo.

app/services/data/get-capacity-breakdown.js
tsheppard01: For national the view which retrieves the capacity graph data (`00104_national_capacity_view.js`) does not insert an id.  So id is set as undefined in the view for national.  
Should I change the graph view so it retrieves an id and change this behaviour?

test/unit/services/test-get-capacity-view.js
```diff
@@ -13,64 +13,155 @@ const CAPACITY_RESULTS = [
...
 })
 
 describe('services/get-capacity-view', function () {
   it('should return a result object with a table, title and breadcrumbs object for ldu', function () {
-    getCapacityStub.resolves(CAPACITY_RESULTS)
+    getWorkloadReports.resolves(CAPACITY_RESULTS)
+    getCapacityBreakdown.resolves(LDU_RESULTS)
 
-    return getCapacityView(5, capacityDateRange, 'ldu').then((result) => {
+    return getCapacityView(callingId, capacityDateRange, 'ldu').then((result) => {
```
andrew-js-wright: 👍

app/services/data/get-capacity-breakdown.js
andrew-js-wright: aah right, I was thinking this was querying each child record individually, in other words - I wasn't really thinking. That makes sense, you want all of the records in the national level's view and therefore don't need to filter. Ignore me :)

tsheppard01: I think most of our views don't insert an id at national level, a work around would be to add `select 0 AS id` to all national views, then we could just pass the url id all the way down.

andrew-js-wright: That sounds like a good solution and it would be nice to have a consistent way of accessing each organisational level but for now this is how it's done everywhere else so I would leave it.

app/views/capacity.html
tsheppard01: It sure is, fixed

PR #133, opened 2017-09-20
Feature/wmt 1013/add date range to capacity view

andrew-js-wright:


app/views/capacity.html
```diff
@@ -6,11 +6,81 @@
...
+                    <label class="form-label" for="capacity-to-year">Year</label>
+                    <input class="form-control" id="capacity-to-year"
+                        name="capacity-to-year" type="number" pattern="[0-9]*"
+                        min="0" max="9999" value={{query['capacity-to-year'] | int}}>
+                </div>
+            </div>
+        </fieldset>
+    </div>
+
+    <input id="caseload-filter-submit" type="submit" value="filter" class="button">
```
caoimheb: Capital on first letter of button text, i.e. Filter?

app/services/validators/validation-error-messages.js
```diff
@@ -1,11 +1,13 @@
...
   getIsRequiredMessage: function (displayName) { return `${displayName} is required` },
   getIsAlphaMessage: function (displayName) { return `${displayName} must only contain letters` },
   getIsNumericMessage: function (displayName) { return `${displayName} must only contain numbers` },
   getIsRangeMessage: function (displayName, options) { return `${displayName} must be between ${options.min} and ${options.max}` },
   getIsIntegerMessage: function (displayName, options) { return `${displayName} must be a whole number between ${options.min} and ${options.max}` },
   getIsFloatMessage: function (displayName, options) { return `${displayName} must be a number between ${options.min} and ${options.max}` },
   getInvalidDateFormatMessage: function (displayName) { return `${displayName} was invalid` },
+  getPastDateMessage: function (displayName) { return `${displayName} must be in the past` },
   getFutureDateMessage: function (displayName) { return `${displayName} must be in the future` },
   getIsDateLaterThanMessage: function (displayName) { return `${displayName} must be after the start date` },
```
caoimheb: If I put in a really old "Show capacity from", e.g. 01 01 1000 the message is "Show capacity from must be after the start date"

If I put in two current dates that are the same, e.g. both 01 09 2017 the message is "Show capacity to must be after the start date"

This is slightly confusing as to what "start date" is. Is it the "Show capacity from" date, or the date at which we started having data?

app/views/capacity.html
```diff
@@ -6,11 +6,81 @@
...
+                <div class="form-group form-group-day">
+                    <label class="form-label" for="capacity-to-day">Day</label>
+                    <input class="form-control" id="capacity-to-day"
+                        name="capacity-to-day" type="number" pattern="[0-9]*"
+                        min="0" max="31" aria-describedby="capacity-to-hint"
+                        value={{query['capacity-to-day'] | int}}>
+                </div>
+                <div class="form-group form-group-month">
+                    <label class="form-label" for="capacity-to-month">Month</label>
+                    <input class="form-control" id="capacity-to-month" name="capacity-to-month" type="number" pattern="[0-9]*" min="0" max="12"
```
caoimheb: We should probably remove the client side validation (min / max) on these inputs and stick with server side, for consistency and use with screen readers.

app/services/validators/validation-error-messages.js
```diff
@@ -1,11 +1,13 @@
 module.exports = {
+  getIsOlderThanMaxHistory: function (displayName, options) { return `Date range must be within the past ${options.years} years` },
   getIsRequiredMessage: function (displayName) { return `${displayName} is required` },
   getIsAlphaMessage: function (displayName) { return `${displayName} must only contain letters` },
   getIsNumericMessage: function (displayName) { return `${displayName} must only contain numbers` },
   getIsRangeMessage: function (displayName, options) { return `${displayName} must be between ${options.min} and ${options.max}` },
   getIsIntegerMessage: function (displayName, options) { return `${displayName} must be a whole number between ${options.min} and ${options.max}` },
   getIsFloatMessage: function (displayName, options) { return `${displayName} must be a number between ${options.min} and ${options.max}` },
   getInvalidDateFormatMessage: function (displayName) { return `${displayName} was invalid` },
+  getPastDateMessage: function (displayName) { return `${displayName} must be in the past` },
```
caoimheb: Input of "31 02 2016", i.e. a date that doesn't exist, gives error of "Show capacity from must be in the past"

test/unit/services/domain/test-capacity-date-range.js
```diff
@@ -37,4 +44,164 @@ describe('services/domain/capacity-date-range', function () {
...
+    expect(function () {
+      return new CapacityDateRange(
+        INVALID_DAY,
+        VALID_FROM_MONTH,
+        VALID_FROM_YEAR,
+        VALID_TO_DAY,
+        VALID_TO_MONTH,
+        VALID_TO_YEAR
+      )
+    }).to.throw(ValidationError)
```
caoimheb: Worth verifying that it is the specific validation error message that we expect?

app/routes/capacity-route.js
```diff
@@ -2,11 +2,25 @@ const getCapacityView = require('../services/get-capacity-view')
...
-    var capacityDateRange = dateRangeHelper.createCapacityDateRange(req.query)
+    var capacityDateRange
+    var errors
+
+    try {
+      capacityDateRange = dateRangeHelper.createCapacityDateRange(req.query)
+    } catch (error) {
+      if (error instanceof ValidationError) {
+        errors = error.validationErrors
+        capacityDateRange = dateRangeHelper.createCapacityDateRange({})
```
ddebella: Where is this capacityDateRange being used. I see its declaration and assignment don't see where its been used?

app/services/domain/capacity-date-range.js
```diff
@@ -14,8 +18,31 @@ class CapacityDateRange {
...
+      .isRequired()
+      .isValidDate()
+      .isPastDate()
+      .isLaterThan(moment().subtract(CASELOAD_CAPACITY.MAX_HISTORY, 'years'), 'maxCapacityHistory')
+      .getFormattedDate()
+
+    this.capacityToDate = FieldSetValidator(this.toFields, 'capacityToDate', errors)
+      .isRequired()
+      .isValidDate(this.capacityToDate)
+      .isPastDate(this.capacityToDate)
```
ddebella: Can the two date not be today's date?

app/services/validators/fieldset-validator.js
```diff
@@ -61,7 +79,11 @@ function isDateInTheFuture (date) {
...
 }
 
-function isDateAfter (startDate, endDate) {
+function isDateInThePast (date) {
+  return validateDate(date) && date < dateFormatter.now()
+}
+
+function isDateAfter (endDate, startDate) {
   return endDate > startDate
 }
```
ddebella: 👍 good additions

test/unit/services/domain/test-capacity-date-range.js
```diff
@@ -1,16 +1,25 @@
...
 const CapacityDateRange = require('../../../../app/services/domain/capacity-date-range')
+const CASELOAD_CAPACITY = require('../../../../app/constants/caseload-capacity')
 
 var capacityDateRange
 
 describe('services/domain/capacity-date-range', function () {
   const VALID_FROM_DAY = '01'
   const VALID_FROM_MONTH = '01'
-  const VALID_FROM_YEAR = '2017'
+  const VALID_FROM_YEAR = (new Date()).getFullYear() - 2
```
ddebella: Should this be some sort of configuration file rather then static `2` years back? If we are sure if it will never change then thats ok.

app/routes/capacity-route.js
andrew-js-wright: It's line 30. It was already there before this PR, hence it's not in the diff.

app/services/domain/capacity-date-range.js
andrew-js-wright: Good point, added `isPastOrPresentDate` to use instead.

app/routes/capacity-route.js
ddebella: Ok

PR #138, opened 2017-10-02
Update capacity and reductions to account for CMS now being an adjust…

caoimheb:
…ment


app/services/data/get-reductions.js
```diff
@@ -6,7 +6,6 @@ module.exports = function (workloadOwnerId) {
   if (workloadOwnerId !== undefined) {
     whereObject.workload_owner_id = workloadOwnerId
   }
-  whereObject.contact_id = null
```
tswann: :+1:

app/services/get-capacity-view.js
```diff
@@ -74,12 +74,18 @@ var parseCapacityBreakdown = function (workloadReports, organisationLevel) {
 }
 
 var buildCapacityBreakdownEntry = function (workloadReport) {
+  var cmsPercentageValue = null
+
+  if (workloadReport.cmsAdjustmentPoints > 0) {
```
tswann: 🤔  What if we have a net negative adjustments total? Could happen if someone is the recipient of any `-points` values.

caoimheb: Myself, Terence and Andy had a chat about this. The decision was that if an OM has had points removed via CMS, then none of their remaining total_points are due to CMS work. So the percentage of their capacity which is being taken up by CMS is 0%. Thoughts?

tswann: Makes sense 👍

PR #141, opened 2017-10-03
Feature/wmt 1039/store logged in user in workload points

ddebella:


test/integration/services/data/test-get-workload-points.js
```diff
@@ -54,6 +54,7 @@ describe('services/data/get-workload-points', function () {
     return getExistingActiveWorkloadPoints().then(function (results) {
       expectedWorkloadPoints = results
       if (results === undefined) {
+        console.log('----------')
```
andrew-js-wright: Remove console.log

app/views/workload-points.html
```diff
@@ -16,9 +16,7 @@
 
 <p>Workload points are set here and are used to calculate capacity points for all Offender Managers throughout the Workload Management Tool.</p>
 
-<!-- TODO: Include user. -->
-
-<p>Current Workload Points created by <span class="bold">test user</span> on <span class="bold">{{ wp.effectiveFrom }}</span></p> 
+<p>Current Workload Points created by <span class="bold">{{ wp.updatedByUserId }}</span> on <span class="bold">{{ wp.effectiveFrom }}</span></p>
```
andrew-js-wright: As mentioned above, this should be a name by the time it's rendered so we could update the property name.

app/services/user-role-service.js
```diff
@@ -56,9 +63,11 @@ module.exports.removeUserByUsername = function (username) {
 }
 
 var removeDomainFromUsername = function (username) {
```
andrew-js-wright: We could expose this as a public method, unit test it and use it in `app/authentication`.

app/authentication.js
```diff
@@ -20,19 +21,23 @@ module.exports = function (app) {
...
     // Get the role for the user
-    getUserRoleByUsername(username).then(function (role) {
-      done(null, {
-        name: displayName,
-        username: username,
-        user_role: role.role,
-        nameID: nameID,
-        nameIDFormat: nameIDFormat
+    getUserByUsername(username).then(function (user) {
+      getUserRoleByUsername(username).then(function (role) {
```
andrew-js-wright: We could use the methods provided by the userRolesService for getting the user and role.

app/services/workload-points-service.js
```diff
@@ -19,11 +20,17 @@ module.exports.getWorkloadPoints = function (id, organisationLevel) {
...
     }
-    result.title = breadcrumbs[0].title
-    result.subTitle = breadcrumbs[1].title
-    result.workloadPoints = results
-    result.breadcrumbs = breadcrumbs
-    return result
+    return userRoleService.getUserById(results.updatedByUserId)
+    .then(function (user) {
+      if (user !== undefined) {
+        results.updatedByUserId = user.name
```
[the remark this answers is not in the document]
ddebella: Regarding the point about we are storing the name, we are storing the user_id into the workload points table. But when that is rendered to the view, we pass the actual name. I agree the `results.updatedByUserID = name` should be changed to something like `results.updatedBy = name`.

app/services/data/get-user-by-id.js
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = function (id) {
+  if (id === undefined) {
+    throw new ReferenceError('User id is not defined')
+  }
+
+  return knex('users')
+    .where('id', id)
+    .select('id',
+            'username',
```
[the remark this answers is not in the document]
ddebella: Done.

app/views/workload-points.html
ddebella: Addressed above

app/services/user-role-service.js
ddebella: done

app/authentication.js
ddebella: Done

app/services/workload-points-service.js
```diff
@@ -14,16 +15,27 @@ module.exports.getWorkloadPoints = function (id, organisationLevel) {
...
-    result.breadcrumbs = breadcrumbs
-    return result
+    return userRoleService.getUserById(workloadPoints.updatedByUserId)
+    .then(function (user) {
+      var updatedBy = workloadPoints.updatedByUserId // Default to the user id
+      if (user) {
+        if (user.name) { // If there is a valid use that
+          updatedBy = user.name
+        }
+        updatedBy = user.username
```
andrew-js-wright: There needs to be an else here or the name will always get over written by the username.

PR #145, opened 2017-10-04
1057: Access right based on user role

ddebella:


app/authorisation.js
```diff
@@ -1,6 +1,20 @@
 const config = require('../config')
 
-var isAuthenticated = function (req) {
+var isUserAuthenticated = function (req) {
+  var result = false
+  if (isAuthenticationEnabled()) {
+    if (!req.isAuthenticated() || !req.user) {
```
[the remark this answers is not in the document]
andrew-js-wright: An example of this type of check can be seen in APVS (the predecessor project to WMT) - https://github.com/ministryofjustice/apvs-internal-web/blob/develop/app/services/authorisation.js

app/routes/admin.js
```diff
@@ -143,3 +146,16 @@ var isValidUsername = function (username) {
...
 }
+
+const validRole = function (req) {
+  if (authorisation.hasRole(req, roles.DATA_ADMIN) || authorisation.hasRole(req, roles.SYSTEM_ADMIN)) {
+    return true
+  }
+  return false
+}
+
+const denieAccess = function (res) {
```
andrew-js-wright: Typo in denyAccess.

app/authorisation.js
```diff
@@ -1,6 +1,20 @@
...
+      result = false
+    } else {
+      result = true
+    }
+  } else {
+    result = true
+  }
+  return result
```
andrew-js-wright: From what I can see, everywhere we use `isUserAuthenticated`, we redirect them to login. This method could do the redirection and we would remove the duplication, which would make it easier to change if required. The method name should be renamed in that case to something like: `assertUserAuthenticated`.

app/routes/contracted-hours.js
```diff
@@ -83,3 +94,13 @@ module.exports = function (router) {
     }
   }
 }
+
+const validRole = function (req) {
+  return authorisation.hasRole(req, roles.OFFENDER_MANAGER)
```
andrew-js-wright: You could update `authorisation.hasRole` to accept an array of roles and an error message. It would then, ensure the user is authenticated, check if the user has any of the roles specified and call `accessDenied` internally.

This would remove duplication and allow you to replace:
```
if (!authorisation.isUserAuthenticated(req)) {
  return res.redirect('/login')
}
if (!validRole(req) { return denieAccess(res) }
```
with:
```
authorisation.hasRole(req, [roles.OFFENDER_MANAGER])
```
and remove all of the `validRole` and `denieAccess` methods in the admin routes.

app/constants/messages.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  UNAUTHORIZED: 'Unauthorized',
+  FORBIDDEN: 'Forbiden',
```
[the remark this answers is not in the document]
ddebella: Done

app/routes/overview.js
```diff
@@ -43,6 +46,8 @@ module.exports = function (router) {
   })
 
   router.get('/:organisationLevel/:id/overview/csv', function (req, res, next) {
+    iauthorisation.assertUserAuthenticated(req, res)
```
tsheppard01: iauthorisation? typo?

app/routes_no_csrf/authentication-saml.js
```diff
@@ -3,6 +3,7 @@ const logger = require('../logger')
 
 module.exports = function (router) {
   router.get('/login', passport.authenticate('saml'), function (req, res) {
+    console.log('----------')
```
tsheppard01: stray? needed?

app/routes/admin.js
```diff
@@ -1,39 +1,59 @@
...
-        userRole = roles.SYSTEM_ADMIN
-      } else {
-        res.status(403)
-        return res.render('includes/error', {
-          error: 'Admin roles required'
+      authorisation.assertUserAuthenticated(req)
+      authorisation.hasRole(req, [roles.SYSTEM_ADMIN, roles.DATA_ADMIN])
+    } catch (error) {
+      if (error instanceof Unathorized) {
+        console.log(error)
```
tsheppard01: If we want to log this error we should use the logger.

PR #146, opened 2017-10-05
Update web to create WPC tasks using staging id instead of workload id

caoimheb:


caoimheb:
Ready for review, but don't merge until `probation-rules` and `worker` are also good to go

PR #147, opened 2017-10-09
Adds page title of form subtitle - title - WMT to all pages

tsheppard01:


app/views/includes/page-title.html
```diff
@@ -0,0 +1,3 @@
+{% if subTitle %}{{ subTitle }} - {% endif %}
+{% if title %}{{ title }}{% endif %} 
+- WMT
```
caoimheb: Maybe include the `-` in the second `if`. Then if there is no subtitle or title the heading won't be `- WMT`, but `WMT`?

caoimheb: We not including any detail about the tab they are on, e.g. `National - HMPPS - Capacity`?

tsheppard01: Andy answered this F2F

tsheppard01: Good spot, fixed

PR #148, opened 2017-10-10
1180: When loging in ivew last page visited and handle bookmark links

ddebella:


app/app.js
```diff
@@ -44,16 +43,6 @@ app.use(favicon(path.join(__dirname, 'govuk_modules', 'govuk_template', 'images'
 
 // Cookie session
```
andrew-js-wright: Can remove this comment now.

app/authorisation.js
```diff
@@ -5,6 +5,11 @@ const Forbidden = require('./services/errors/authentication-error').Forbidden
 var assertUserAuthenticated = function (req) {
   if (isAuthenticationEnabled()) {
     if (!req.user) {
+      // Check if the request is coming from a valid user and
+      // store their last request. This is used to handle bookmark links
```
andrew-js-wright: This comment is a little misleading. This will store unauthenticated requests only.

app/routes_no_csrf/authentication-saml.js
```diff
@@ -3,21 +3,29 @@ const logger = require('../logger')
...
+    setRedirectTo(req.session, req.headers.referer)
     passport.logout(req, res)
     if (req.user) {
       logger.info(req.user.nameID, 'logged out.')
     }
     req.logout()
   })
 }
+
+var setRedirectTo = function (session, redirectTo) {
```
andrew-js-wright: Why do we need this function? We check the `redirectTo` variable is checked before read on line 14. If the `redirectTo` variable is undefined we will fallback to the default address.

ddebella: The function is used in the get request for logout and login routes. For the post route we don't don't need to do any set request call hence we only needed to check if the redirect is valid or not.

app/app.js
ddebella: Done

app/authorisation.js
ddebella: Updated the comment.

app/routes_no_csrf/authentication-saml.js
andrew-js-wright: Sorry, I didn't explain myself very well. I meant that we didn't need to worry about writing a `undefined` value to the session variable because we're handling it on read. However, thinking about it, it makes sense that if there is already a `redirectTo` field set we probably shouldn't overwrite it with undefined so happy for this to stay. Thanks.

PR #156, opened 2017-10-16
1212: Remove getContractedHours as it references redundant fields.

andrew-js-wright:
The application was giving regions with 0 contracted hours e.g. `Stakeholder
Engagement London` the default number of contracted hours for a PO or PSO. There
are some instances however, where the number of contracted hours should indeed
be 0. This is when migrated data has been brought over and there are regions
with no LDUs associated with them.

### DIFF
--- a/test/unit/services/test-get-overview.js
+++ b/test/unit/services/test-get-overview.js
@@ -17,30 +17,14 @@ const OVERVIEW = {
   cases: '2',
   hours: '3',
   reduction: '4',
-  contractedHours: 37,
-  defaultContractedHoursPo: 3,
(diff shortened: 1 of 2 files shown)


test/unit/services/test-get-overview.js
```diff
@@ -144,4 +128,14 @@ describe('services/get-overview', function () {
       expect(result.overviewDetails).to.eql(ZERO_AVAILABLE_POINTS_OVERVIEWS)
     })
   })
+
+  it('should 0 contracted hours if there are indeed 0 contracted hours', function () {
```
caoimheb: Minor: Word missing between should and 0?

PR #158, opened 2017-10-16
414: Add arms case count to the web views

andrew-js-wright:


app/views/capacity.html
```diff
@@ -137,7 +137,7 @@ <h3 class="heading-medium">Capacity breakdown</h3>
           <td headers="paroms breakdown">-</td>
           <td headers="fdrs breakdown">-</td>
           <td headers="sdrs breakdown">-</td>
-          <td headers="arms breakdown">-</td>
+          <td headers="arms breakdown">{{ member.armsTotalCases }}</td>
```
ddebella: should this not be rounded too? `{{ member.armsTotalCases | round(1) }}`

app/views/capacity.html
```diff
@@ -158,7 +158,7 @@ <h3 class="heading-medium">Capacity breakdown</h3>
           <td headers="paroms breakdown">-</td>
           <td headers="fdrs breakdown">-</td>
           <td headers="sdrs breakdown">-</td>
-          <td headers="arms breakdown">-</td>
+          <td headers="arms breakdown">{{ member.grades[i].armsTotalCases }}</td>
```
ddebella: Same as above

app/views/capacity.html
andrew-js-wright: This is an integer and therefore shouldn't need to be rounded.

app/views/capacity.html
andrew-js-wright: Same as above.

andrew-js-wright:
Looking into the failing tests now.

PR #159, opened 2017-10-16
Fix/materialise views

caoimheb:


knex.js
```diff
@@ -1,4 +1,12 @@
-const knexWebSchema = require('knex')(require('./knexfile').web)
+var config = require('./knexfile').web
+
+config.pool.afterCreate = function (conn, done) {
+  conn.batch(['SET ARITHABORT ON'])
+  .then((err, res) => done(err, conn))
+  .catch((err, res) => done(err, conn))
+}
+
+const knexWebSchema = require('knex')(config)
```
ddebella: 👍

andrew-js-wright:
This was timing out when running integration tests because the after hook was taking 47 seconds. This has been fixed in commit d29bb94.

PR #166, opened 2017-10-19
Feature/wmt 1003/capacity red

davidmck95:
https://tree.taiga.io/project/chagarwa-workload-management-tool/issue/1003


app/assets/sass/_wmt/_wmt-data-table.scss
```diff
@@ -54,6 +54,10 @@ table.data-table tr.data-sub-header td:nth-child(7) {
     font-weight: bold;
 }
 
+table td.capacityRed {
```
andrew-js-wright: Minor: in our css class names we usually put a hyphen between the words rather than using camel case, so this would become `capacity-red`.

PR #171, opened 2017-10-23
Feature/wmt 1002/round graph figures

davidmck95:
Rounding graph figures that appear when mouse hovers over. 
Anything greater than or equal to 0.6% round it upwards to the nearest whole number, and anything below 0.6%, round it downwards to the nearest whole number.

### DIFF
--- a/app/services/helpers/custom-round.js
+++ b/app/services/helpers/custom-round.js
@@ -1,4 +1,4 @@
-module.exports.customRound = function (number) {
+module.exports = function (number) {
   if (number || number === 0) {
     if ((number % 1) !== 0) {
       let decimal = (number % 1).toFixed(4)
--- a/app/services/helpers/custom-round.js
+++ b/app/services/helpers/custom-round.js
@@ -1,0 +1,15 @@
+module.exports.customRound = function (number) {
+  if (number || number === 0) {
+    if ((number % 1) !== 0) {
+      let decimal = number % 1
+      if (decimal < 0.5999999999999943 && decimal >= 0.5) {
+        number = Math.floor(number)
+      } else {
+        number = Math.round(number)
+      }
+    }
+    return number
+  } else {
+    return undefined
+  }
+}
--- a/app/services/helpers/custom-round.js
+++ b/app/services/helpers/custom-round.js
@@ -2,7 +2,7 @@ module.exports.customRound = function (number) {
   if (number || number === 0) {
     if ((number % 1) !== 0) {
       let decimal = number % 1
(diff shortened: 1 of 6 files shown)


app/services/helpers/table-creator.js
```diff
@@ -30,3 +30,13 @@ module.exports.createCapacityTable = function (id, rowLabel, capacityDateRange,
 
   return new DisplayTable(headings, rows)
 }
+
+function customRound (number) {
```
andrew-js-wright: Some unit tests on this would be nice - Like what happens if the `number` is undefined?

davidmck95:
Found bug - working on fix

davidmck95:
Using toFixed when finding the decimal part of the number helps to avoid rounding problems.
JS: 155.6 - 155 = 0.5999999999999943
14.6 % 1 = 0.5999999999999996

app/services/helpers/custom-round.js
```diff
@@ -0,0 +1,15 @@
+module.exports.customRound = function (number) {
```
caoimheb: Since this file only has one function you can use `module.exports = function (number) {` and then when you `require` it, the import acts as the function, ie. 

`const customRound = require('./custom-round')`
`capacityPercentageRow.values.push(customRound(capacity['capacity_percentage']))`

test/unit/services/helpers/test-reduction-hours-calculator.js
```diff
@@ -27,7 +27,7 @@ var reductionWithHoursResults = [
 var woContractedHours = 37
 
 describe('services/helpers/reduction-hours-calculator', function () {
-  it('should popluate absolute values of reduction allowance hours, for each reason, for the particular offender manager', function () {
+  it('should populate absolute values of reduction allowance hours, for each reason, for the particular offender manager', function () {
```
caoimheb: 👍

test/unit/services/helpers/test-custom-round.js
```diff
@@ -0,0 +1,33 @@
+const expect = require('chai').expect
+const tableCreator = require('../../../../app/services/helpers/custom-round')
```
caoimheb: Update function name as in other comment

app/views/capacity.html
```diff
@@ -91,13 +91,13 @@
         <tr>
             <th scope="row">Capacity</th>
             {% for i in range(0, capacity.rows[0].values.length)%}
-                <td> {{capacity.rows[0].values[i] | round}}% </td>
+                <td> {{capacity.rows[0].values[i]}} </td>
```
caoimheb: `%` still needed

PR #173, opened 2017-10-23
Feature/wmt 1225/export caseload info

c-mcateer:
Added in a LDU Cluster column in the exported CSV file for teams and offender manager


app/services/get-export-csv.js
```diff
@@ -43,12 +42,14 @@ var getFields = function (organisationLevel, tab) {
...
         childOrgForFieldName = getChildOrgForFieldName(organisationLevel)
-        fields = Object.assign([], ORG_OVERVIEW_FEILDS)
+        fields = Object.assign([], ORG_OVERVIEW_FIELDS)
         fieldNames = [childOrgForFieldName + 'Name', 'CapacityPercentage', 'CapacityPoints', 'ContractedHours', 'ReductionHours', 'TotalCases']
 
         if (organisationLevel === organisationUnitConstants.TEAM.name) {
           fields.push('gradeCode')
           fieldNames.push('GradeCode')
+          fields.unshift('lduCluster')
+          fieldNames.unshift('LDUCluster')
```
andrew-js-wright: According to the screenshots in the story, I think the team name should also be in the outputted CSV when at the team level. This may have changed though.

andrew-js-wright: Also, the names for the columns should have spaces between them. So `LDUCluster` should be `LDU Cluster`.

PR #175, opened 2017-10-23
Ensure get-capacity-breakdown uses index from view

andrew-js-wright:


app/services/data/get-capacity-breakdown.js
```diff
@@ -1,30 +1,54 @@
...
+    'cms_adjustment_points AS cmsAdjustmentPoints',
+    'gs_adjustment_points AS gsAdjustmentPoints',
+    'arms_total_cases AS armsTotalCases',
+    'contracted_hours AS contractedHours'
+  ]
+
+  var requiresWorkloadOwnerName = (type === ORGANISATION_UNIT.TEAM.name)
+
+  if (requiresWorkloadOwnerName) {
+    selectList.push('forename')
```
caoimheb: Could you push `CONCAT(forename, ' ', surname) AS name`, and then the loop processing after the query wouldn't be required?

andrew-js-wright: Great shout, cheers.

PR #176, opened 2017-10-23
Feature/wmt 996/t2a workload point calculation

ddebella:


app/routes/workload-points.js
```diff
@@ -33,7 +34,39 @@ module.exports = function (router) {
           breadcrumbs: result.breadcrumbs,
           wp: result.workloadPoints,
           updatedBy: result.updatedBy,
-          successText: successText
+          successText: successText,
+          isT2A: false
```
dconey646: Should this always be false?

app/routes/workload-points.js
```diff
@@ -33,7 +34,39 @@ module.exports = function (router) {
...
+      } else if (error instanceof Forbidden) {
+        return res.status(error.statusCode).render(error.redirect, {
+          heading: messages.ACCESS_DENIED,
+          message: messages.ADMIN_ROLES_REQUIRED
+        })
+      }
+    }
+    var success = req.query.success
+    var successText = success ? 'You have successfully updated the workload points for transition to adulthood cases!' : null
+    var isT2A = true
```
dconey646: Var isn't needed if we are setting it below

test/unit/services/domain/test-workload-points.js
```diff
@@ -77,6 +78,7 @@ describe('services/domain/workload-points', function () {
     expect(workloadPoints.weightingArmsCommunity).to.equal('10.1')
     expect(workloadPoints.weightingArmsLicense).to.equal('10.2')
     expect(workloadPoints.weightingWarrants).to.equal('0.0')
+    expect(workloadPoints.isT2A).to.equal('false')
   })
 
   it('should raise a ValidationError if a Community Weighting field is out of range', function () {
```
dconey646: Could we add a unit test to test if the isT2A flag is set?

app/routes/workload-points.js
ddebella: Agreed.

app/routes/workload-points.js
ddebella: This route is specifically for non-T2A workload points and so it should always be false.

app/views/workload-points.html
```diff
@@ -471,6 +475,18 @@
               </table>
             </div>
           </div>
+          {% endif %}
+          {% if wp.isT2A %}
+          <input class="form-control" id="sdr" name="sdr" type="hidden" value="{{ wp.sdr }}">
```
caoimheb: These values aren't needed for T2A? Do they need to be included here? What happens if they are removed - will the form still submit but with `null`s?

package.json
```diff
@@ -86,6 +86,6 @@
     "underscore": "^1.8.3",
     "util": "^0.10.3",
     "validator": "^8.2.0",
-    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#0.1.8"
+    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#0.5.0"
```
caoimheb: Tag should include `v`

test/unit/routes/test-workload-points.js
```diff
@@ -87,6 +89,10 @@ describe('Admin Workload Points route', function () {
     return supertest(app).get(WORKLOAD_POINTS_URL).expect(200)
   })
 
+  it('should respond with 200 when the correct admin/workload-points/t2s url is called', function () {
```
caoimheb: Typo - `t2a`

test/unit/routes/test-workload-points.js
```diff
@@ -87,6 +89,10 @@ describe('Admin Workload Points route', function () {
...
+    return supertest(app).get(WORKLOAD_POINTS_T2A_URL).expect(200)
+  })
```
caoimheb: Can you add to these tests to check that T2A url calls the service with `t2a=true` and regular calls the service with `t2a=false`

app/services/validators/field-validator.js
```diff
@@ -56,6 +56,13 @@ class FieldValidator {
     }
     return this
   }
+
+  isBoolean () {
+    if (this.data && !validator.isBoolean(this.data.toString())) {
+      this.errors.add(this.fieldName, ERROR_MESSAGES.getIsBooleanMessage)
+    }
+    return this
+  }
```
caoimheb: Add unit tests for this?

test/unit/services/domain/test-workload-points.js
ddebella: 👍

app/views/workload-points.html
ddebella: If these values aren't passed in the the workload validator fails the validation and the object isn't created. We could have a separate workload object for T2A. Given that both the standard and t2a are using the same table its make sense to have a single workload point object. Another possibility would not to validate these values for T2A points based on the isT2A flag.

package.json
ddebella: good catch.

package.json
```diff
@@ -86,6 +86,6 @@
     "underscore": "^1.8.3",
     "util": "^0.10.3",
     "validator": "^8.2.0",
-    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#0.5.0"
+    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#^0.5.0"
```
caoimheb: Typo - `v0.5.0`

test/unit/services/validators/test-field-validator.js
```diff
@@ -221,4 +221,24 @@ describe('services/validators/field-validator', function () {
...
+    it('should return false if passed in a valid false value', function () {
+      FieldValidator('false', FIELD_NAME, errorHandler).isBoolean()
+      var errors = errorHandler.get()
+      expect(errors).to.equal(false)
+    })
+
+    it('should not return false if passed in invalid true or false value', function () {
+      FieldValidator('xxxx', FIELD_NAME, errorHandler).isBoolean()
+      var errors = errorHandler.get()
+      expect(errors).to.not.equal(true)
```
caoimheb: If `should not return false`, then why why checking for `not.equal(true)`?

test/unit/services/domain/test-workload-points.js
```diff
@@ -81,6 +81,55 @@ describe('services/domain/workload-points', function () {
     expect(workloadPoints.isT2A).to.equal('false')
   })
 
+  it('should create a valid T2A workload points object', function () {
+    var t2aWorkloadPoints = new WorkloadPoints(Object.assign({}, VALID_WORKLOAD_POINTS))
+    t2aWorkloadPoints.isT2A = 'true'
+    t2aWorkloadPoints.defaultContractedHoursPo = '0'
```
caoimheb: If you're assigning the values here and checking the below, it's not really testing the `WorkloadPoints` function. I'm not sure this test is needed

test/unit/routes/test-workload-points.js
```diff
@@ -89,19 +120,33 @@ describe('Admin Workload Points route', function () {
...
+  it('should respond with 500 when an incorrect url is called for t2a', function () {
+    return supertest(app).get(INVALID_WORKLOAD_POINTS_T2A_URL).expect(500)
+  })
+
+  it('should respond with 200 when the correct post admin/workload-points/t2a url is called', function () {
+    return supertest(app)
+      .post(WORKLOAD_POINTS_T2A_URL)
+      .send(MOCK_USER)
+      .send(WORKLOAD_POINTS_T2A_TO_POST)
+      .expect(302, 'Found. Redirecting to /admin/workload-points/t2a?success=true')
```
caoimheb: Checks that the service is called with the correct `true` / `false` param?

test/unit/services/domain/test-workload-points.js
ddebella: I agree, removed the test.

PR #177, opened 2017-10-23
Format capacity percentage figure being exported to csv

davidmck95:
Formats capacity percentage from overview table to two decimal places and appends a percentage sign to the end.

### DIFF
--- a/app/services/get-export-csv.js
+++ b/app/services/get-export-csv.js
@@ -89,9 +89,13 @@ var getCsv = function (organisationLevel, result, tab, fields, fieldNames) {
       }
       break
     case tabs.OVERVIEW:
-      result.overviewDetails.forEach(function(item) {
-        item.capacityPercentage = parseFloat(item.capacityPercentage).toFixed(2) + '%'
-      })
+      if(Array.isArray(result.overviewDetails)) {
+        result.overviewDetails.forEach(function(item) {
+          item.capacityPercentage = formatCapacityValue(item.capacityPercentage)
+        })
+      } else {
+        result.overviewDetails.capacity = formatCapacityValue(result.overviewDetails.capacity)
+      }
       csv = generateCsv(result.overviewDetails, fields, fieldNames)
       break
   }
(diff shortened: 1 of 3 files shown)


andrew-js-wright:
Checks are failing, a ‘standard —fix’ should sort it

test/unit/services/test-get-export-csv.js
```diff
@@ -36,4 +36,20 @@ describe('services/get-export-csv', function () {
       expect(getExportCsv(orgUnit.NATIONAL.name, helper.LDU_OVERVIEW_RESULT, tabs.OVERVIEW).csv).to.include(helper.NATIONAL_OVERVIEW_HEADINGS)
     })
   })
+  describe('should format the capacity when exporting overviews', function () {
+    it('should be formatted to two decimal figures', function () {
```
caoimheb: Minor: Maybe rename tests here. This should read like a sentence - outer describe, inner describe, it. So in this case:
`services/get-export-csv... should format the capacity when exporting overviews... should be formatted to two decimal figures`
should be something like 
`services/get-export-csv... should format the capacity when exporting overviews... to two decimal figures`

app/services/get-export-csv.js
```diff
@@ -89,6 +89,13 @@ var getCsv = function (organisationLevel, result, tab, fields, fieldNames) {
       }
       break
     case tabs.OVERVIEW:
+      if (Array.isArray(result.overviewDetails)) {
```
caoimheb: Is this an Array for all levels above offender manager? If so, maybe do the `if` on organisational level to match other `if`s in this file

test/unit/services/test-get-export-csv.js
```diff
@@ -37,17 +37,17 @@ describe('services/get-export-csv', function () {
     })
   })
   describe('should format the capacity when exporting overviews', function () {
-    it('should be formatted to two decimal figures', function () {
+    it('to two decimal figures', function () {
       var capacityExport = getExportCsv(orgUnit.TEAM.name, helper.TEAM_OVERVIEW_RESULT, tabs.OVERVIEW).csv
       expect(capacityExport).to.include('107.37')
```
caoimheb: Since this is an include, it doesn't fully test that it is rounded to 2 decimal places, as a value of `107.37389748329` would still pass here as it includes `107.37`. Including the `%` symbol should sort this

PR #180, opened 2017-10-24
Call indexed case-progress views

andrew-js-wright:
Ensure get-capacity-breakdown and case-progress are called with index.

### DIFF
--- a/app/services/data/get-capacity-breakdown.js
+++ b/app/services/data/get-capacity-breakdown.js
@@ -1,29 +1,45 @@
 const knex = require('../../../knex').web
 const orgUnitFinder = require('../helpers/org-unit-finder')
+const ORGANISATION_UNIT = require('../../constants/organisation-unit')
 
 module.exports = function (id, type) {
   var orgUnit = orgUnitFinder('name', type)
 
   var table = orgUnit.capacityBreakdownView
-  var whereObject = {}
 
-  if (id !== undefined) {
-    whereObject.id = id
+  var selectList = [
+    'total_points AS totalPoints',
+    'available_points AS availablePoints',
+    'reduction_hours AS reductionHours',
+    'link_id AS linkId',
+    'grade_code AS grade',
+    'total_cases AS totalCases',
+    'cms_adjustment_points AS cmsAdjustmentPoints',
+    'gs_adjustment_points AS gsAdjustmentPoints',
+    'arms_total_cases AS armsTotalCases',
+    'contracted_hours AS contractedHours'
+  ]
+
+  var requiresWorkloadOwnerName = (type === ORGANISATION_UNIT.TEAM.name)
+
+  if (requiresWorkloadOwnerName) {
+    selectList.push('CONCAT(forename, \' \', surname) AS name')
+  } else {
+    selectList.push('name')
   }
 
-  return knex(table)
-    .where(whereObject)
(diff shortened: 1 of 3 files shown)


app/services/data/get-capacity-breakdown.js
```diff
@@ -1,29 +1,45 @@
...
+    'cms_adjustment_points AS cmsAdjustmentPoints',
+    'gs_adjustment_points AS gsAdjustmentPoints',
+    'arms_total_cases AS armsTotalCases',
+    'contracted_hours AS contractedHours'
+  ]
+
+  var requiresWorkloadOwnerName = (type === ORGANISATION_UNIT.TEAM.name)
+
+  if (requiresWorkloadOwnerName) {
+    selectList.push('CONCAT(forename, \' \', surname) AS name')
```
ddebella: I think we should be just returning the forename and surname as fields and let the user of this call handle the  concatenation

app/services/data/get-capacity-breakdown.js
```diff
@@ -1,29 +1,45 @@
...
+  } else {
+    selectList.push('name')
+  }
```
ddebella: Not clear on this select list. Is there an attribute called `name` or we have all three attributes, `forename`, `surname` and `name`?
I also think we should be just returning the forename and surname as fields and let the user of this call handle the  concatenation

app/services/data/get-caseload-progress.js
```diff
@@ -1,15 +1,14 @@
...
 module.exports = function (id, type) {
   var orgUnit = orgUnitFinder('name', type)
   var table = orgUnit.caseProgressView
-  var whereObject = {}
+  var whereString = ''
 
-  if (id !== undefined) {
-    whereObject.id = id
+  if (id !== undefined && (!isNaN(parseInt(id, 10)))) {
+    whereString += ' WHERE id = ' + id
```
ddebella: this should be `whereString =` not needed to do `+=`

app/services/data/get-caseload-progress.js
```diff
@@ -21,32 +20,21 @@ module.exports = function (id, type) {
...
-  if (requiresWorkloadOwnerName) {
-    selectList.push('workload_owner_id')
+  var noExpandHint = ''
+
+  if (isIndexed) {
+    selectList.push('CONCAT(forename, \' \', surname) AS name')
+    noExpandHint = ' WITH (NOEXPAND)'
   } else {
     selectList.push('name')
   }
```
ddebella: Same as above

app/services/data/get-capacity-breakdown.js
andrew-js-wright: The reason this is here is because name is handled in different ways depending on the organisation level - if it is a team then you'll want the offender manager names which will have to be concatenated from `om.forname` and `om.surname`. In the case of an ldu, region etc. we can just use the name property on the view.

The reason this is done here and not by the calling class is because the calling class shouldn't need to care about the organisation type, whereas the data access layer needs to be aware of it.

This concatenation used to be done in the view, but in order to apply the indexes a function couldn't be applied to a group by field.

app/services/data/get-caseload-progress.js
andrew-js-wright: I'll change this now.

PR #185, opened 2017-10-25
Feature/wmt 1182/add total rows to caseload tables

dconey646:


app/views/caseload.html
```diff
@@ -165,6 +184,22 @@ <h3 class="heading-medium">{{ caseType.displayName | title }} caseload by tier</
               </tr>
             {% endfor %}
             {% endfor %}
+            <tr>
+                <td rowspan="2" headers="{{caseType.displayName}}_om {{caseType.displayName}}_name">Total</td>
```
andrew-js-wright: I think the rowspan of these will be determined by the number of grades so it should probably by dynamic e.g. `caseType.array.totals.length`.

app/views/caseload.html
```diff
@@ -216,6 +251,22 @@ <h3 class="heading-medium">Caseload by grade</h3>
           </tr>
           {% endfor %}
         {% endfor %}
+        <tr>
+          <td rowspan="2" headers="{{caseType.displayName}}_om">Average</td>
```
andrew-js-wright: Same with this `rowspan`

app/services/helpers/caseload-helper.js
```diff
@@ -131,13 +180,46 @@ var transform = function (caseloadTotalsByGrade, calculatePercentage = false) {
...
+    gradeTotals[newGradeRecord.grade].c1 += newGradeRecord.c1
+    gradeTotals[newGradeRecord.grade].c2 += newGradeRecord.c2
+    gradeTotals[newGradeRecord.grade].d1 += newGradeRecord.d1
+    gradeTotals[newGradeRecord.grade].d2 += newGradeRecord.d2
+    gradeTotals[newGradeRecord.grade].untiered += newGradeRecord.untiered
+    gradeTotals[newGradeRecord.grade].totalCases += newGradeRecord.totalCases
+    gradeTotals[newGradeRecord.grade].numberOfType++
+  } else {
+    gradeTotals[newGradeRecord.grade] = Object.assign({}, newGradeRecord)
+    gradeTotals[newGradeRecord.grade].numberOfType = 1
```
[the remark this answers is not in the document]
andrew-js-wright: Ignore this - after a conversation, I see the `numberOfType` field is different to the `gradeTotals.length`.

app/views/capacity.html
```diff
@@ -130,6 +130,17 @@ <h3 class="heading-medium">Capacity breakdown</h3>
       <tbody>
         {% for member in capacityBreakdown %}
         <tr>
+          {% if member.name === 'Total/Average' %}
```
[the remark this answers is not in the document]
dconey646: Good shout, yea it will always be the last position in the array as we only do the totals after the initial array is built up.

app/views/caseload.html
dconey646: An even better shout, will get that in too.

app/views/caseload.html
dconey646: 👍

app/views/capacity.html
```diff
@@ -174,6 +187,20 @@ <h3 class="heading-medium">Capacity breakdown</h3>
           <td headers="cms breakdown">{{ member.grades[i].cmsPercentage | round(1) }}%</td>
         </tr>
         {% endfor %}
+        {% if loop.last %}
+        <tr>
+            <td colspan="2" headers="name org">{{ member.name }}</td>
+            <td headers="capacity breakdown">{{ member.capacity | round(1) }}%</td>
+            <td headers="cases breakdown">{{ member.totalCases | round(1) }}</td>
+            <td headers="cases breakdown">{{ member.totalTotalT2aCases | round(1) }}</td>
```
caoimheb: Header include `t2aCases` as opposed to `cases`? Here and on line 138

app/services/get-overview.js
```diff
@@ -25,16 +25,28 @@ module.exports = function (id, organisationLevel) {
 }
 
 var calculateValues = function (results) {
+  var totals = { name: 'Total/Average', totalPercentage: 0, totalPoints: 0, totalContractedHours: 0, totalReduction: 0, totalTotalCases: 0 }
+  var thingToReturn = {}
```
caoimheb: Maybe rename to `totalsToReturn` or something a bit more descriptive?

app/services/get-capacity-view.js
```diff
@@ -35,11 +35,17 @@ module.exports = function (id, capacityDateRange, organisationLevel) {
 
 var parseCapacityBreakdown = function (workloadReports, organisationLevel) {
   var capacityBreakdown = []
+  var totals = { name: 'Total/Average', capacity: 0, totalCases: 0, totalARMS: 0, totalGs: 0, totalCMS: 0, totalSDRs: 0, totalParoms: 0, totalSdrConversions: 0, totalTotalT2aCases: 0 }
```
caoimheb: Possibly update to `Total / Average` throughout with spaces?

dconey646: Yep, sure.

app/services/get-overview.js
dconey646: Yea.....was kinda....yea, I don't even have an excuse 😆

app/views/capacity.html
dconey646: Ah balls, good spot. Will get that in.

app/views/includes/organisation-overview.html
```diff
@@ -65,6 +65,17 @@
     <tbody>
     {% for member in overviewDetails %}
       <tr >
+        {% if member.name === 'Total/Average' %}
```
caoimheb: Can we also use something like `{% if loop.last %}` here?

PR #194, opened 2017-10-27
Hotfix for t2a workload points update not taking into account for rec…

ddebella:
…alculation.

### DIFF
--- a/app/services/data/get-ids-for-workload-points-recalc.js
+++ b/app/services/data/get-ids-for-workload-points-recalc.js
@@ -1,6 +1,10 @@
 const knex = require('../../../knex').web
 
-module.exports = function (previousWorkloadPointsId) {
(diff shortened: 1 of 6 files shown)


app/services/data/get-ids-for-workload-points-recalc.js
```diff
@@ -9,7 +13,7 @@ module.exports = function (previousWorkloadPointsId) {
     .max('workload_report.id AS workloadReportId')
     .whereNotNull('workload_report.effective_from')
     .whereNull('workload_report.effective_to')
-    .where('workload_points_calculations.workload_points_id', previousWorkloadPointsId)
+    .where(whereColumnName, previousWorkloadPointsId)
```
caoimheb: Can we add an integration test for this query with `t2a = true`?

PR #196, opened 2017-10-27
Adding remaining points to columns and also removing total from exports

dconey646:


app/services/get-overview.js
```diff
@@ -17,33 +17,37 @@ module.exports = function (id, organisationLevel) {
...
-  var totals = { name: 'Total / Average', totalPercentage: 0, totalPoints: 0, totalContractedHours: 0, totalReduction: 0, totalTotalCases: 0 }
+var calculateValues = function (results, isCSV) {
+  var totals = { name: 'Total / Average', totalPercentage: 0, totalPoints: 0, totalContractedHours: 0, totalReduction: 0, totalTotalCases: 0, totalRemainingPoints: 0 }
   var totalsToReturn = {}
   if (results.length !== undefined) {
     results.forEach(function (result) {
       result.capacityPercentage = 0
       if (result.availablePoints > 0) {
         result.capacityPercentage = (result.totalPoints / result.availablePoints) * 100
+        result.remainingPoints = result.availablePoints - result.totalPoints
```
caoimheb: Think this can go outside the if. If they have zero available points but cases assigned, that should probably be displayed as a negative remaining points

dconey646: Okay, cool. I wasn't too sure. Will do that now.

PR #205, opened 2017-11-03
Export a full list of OMs Caseload figures, hours, reductions & capacity data for each level

davidmck95:


app/services/data/get-full-overview.js
```diff
@@ -0,0 +1,39 @@
+const knex = require('../../../knex').web
+const orgUnitFinder = require('../helpers/org-unit-finder')
+const orgUnitConstants = require('../../constants/organisation-unit')
+
+module.exports = function (id, type) {
+  var orgUnit = orgUnitFinder('name', type)
+  var table = 'individual_case_overview'
+  var whereClause = ''
+  var orderBy = 'ldu_name, team_name'
```
c-mcateer: In the feedback email its says order by organisational level, so if this is at the region level, should it not be ordered by the division name?

app/services/data/get-full-overview.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  return knex.raw(
+        'SELECT ' + selectColumns.join(', ') +
+        ' FROM ' + table + ' WITH (NOEXPAND)' +
+        whereClause + ' ORDER BY ' + orderBy)
+    .then(function (results) {
+      return results
+    })
+}
```
c-mcateer: Add integration testing for the data retrieval

test/helpers/export-csv-helper.js
```diff
@@ -226,84 +226,611 @@ module.exports.OM_OVERVIEW_RESULT = {
...
+  subTitle: 'Team'
 }
 
 module.exports.TEAM_OVERVIEW_CSV = {
-  filename: 'Test_Team_Overview.csv',
-  csv: '"LDU Cluster","Team Name","Offender Manager Name","Capacity Percentage","Capacity Points","Contracted Hours","Reduction Hours","Total Cases","Grade Code"\n' +
-  '"Test Cluster","Test Team","John Smith","107.37%",190,37.5,6,63,"PO"\n' +
-  '"Test Cluster","Test Team","Tony Test","106.84%",190,36.5,6,63,"PO"'
+  filename: 'Team_1_Overview.csv',
+  csv: '"LDU Cluster","Team Name","Offender Manager","Grade Code","Capacity Percentage","Capacity Points","Contracted Hours","Reduction Hours","Total Cases"\n"LDU Cluster 1","Team 1","John Smith","PO","115%",190,37.5,6,69\n"LDU Cluster 1","Team 1","Tony Test","PSO","113%",190,37.5,3,69\n"LDU Cluster 1","Team 1","Jane Doe","PO","116%",190,37.5,1,69\n"LDU Cluster 1","Team 1","Marcin Martin","PSO","107%",190,37.5,6,69\n"LDU Cluster 1","Team 1","Courtney Larry","PO","117%",190,37.5,3,69'
```
c-mcateer: These could be formatted like the test csv files in this file, but putting a '+' after '\n' and the actually taking a new line to make the code clearer and easier to read, instead of it being on really long line

app/services/data/get-full-overview.js
```diff
@@ -0,0 +1,39 @@
...
+    'available_points AS availablePoints',
+    'total_points AS totalPoints',
+    'contracted_hours AS contractedHours',
+    'reduction_hours AS reductionHours',
+    'grade_code AS gradeCode'
+  ]
+
+  if (orgUnit.name === orgUnitConstants.REGION.name || orgUnit.name === orgUnitConstants.NATIONAL.name) {
+    selectColumns.unshift('region_name AS regionName')
+    orderBy = ' region_name, ' + orderBy
```
ddebella: I would remove the leading and trailing spaces on the region_name

app/services/data/get-individual-overview.js
```diff
@@ -12,6 +12,8 @@ module.exports = function (id) {
     'grade_code AS grade',
     'team_id AS teamId',
     'team_name AS teamName',
+    'ldu_name AS lduCluster',
+    'region_name AS region',
```
ddebella: this should be called `regionName` to be consistent with the other levels.

app/services/get-export-csv.js
```diff
@@ -168,5 +155,5 @@ var parseTotalSummaryTable = function (totalSummary) {
 }
 
 var formatCapacityValue = function (capacity) {
-  return parseFloat(capacity).toFixed(2) + '%'
+  return parseInt(capacity) + '%'
```
ddebella: I thought the requirement was that we show two decimal places?

davidmck95: Yeah you're right, I originally wrote that function but when I was getting feedback from Urban about these new export files he seemed to change his mind and asked for it to be a whole number. Below is his feedback from email:

`As far as I can see, the data displayed in the mock exports meets the requirement. I have copied in Andy who raised this point. In line with other feedback provided could the capacity percentage be displayed as whole number percentage, i.e. 97% instead 96.72%.`

Looking back though, I should probably be using Math.round rather than parseInt

app/services/data/get-full-overview.js
davidmck95: This is added on line 29 if the organisation level is region or above

app/services/get-export-csv.js
ddebella: 👍

PR #218, opened 2017-11-28
Fix/button over headers

c-mcateer:
Fixed the export caseload and export reduction button, so that they don't appear on top of the tables in older versions of Firefox.

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -14,13 +14,10 @@ $path: "/public/images/";
   position: absolute !important;
 }
 
-a.reduction-export:visited{
+a.button:visited{
     color:white !important;
 }
 
-a.sln-export:visited{
-    color:white !important;
-}
 @mixin respond-to($media) {
   @if $media==handhelds {
     @media only screen and (max-width: 568px) {
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -14,6 +14,10 @@ $path: "/public/images/";
   position: absolute !important;
 }
 
+
+table.export-buttons td a:visited {
+    color:white !important;
+}
 @mixin respond-to($media) {
   @if $media==handhelds {
     @media only screen and (max-width: 568px) {
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -14,8 +14,11 @@ $path: "/public/images/";
   position: absolute !important;
 }
 
+a.reduction-export:visited{
+    color:white !important;
+}
 
-table.export-buttons td a:visited {
+a.sln-export:visited{
     color:white !important;
 }
 @mixin respond-to($media) {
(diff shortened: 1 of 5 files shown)


app/assets/sass/application.scss
```diff
@@ -14,6 +14,10 @@ $path: "/public/images/";
   position: absolute !important;
 }
 
+
+table.export-buttons td a:visited {
```
davidmck95: Should this class be renamed to export-button and be applied to the <a> element directly.

davidmck95: Does the export button under caseload tab also need this class?

app/views/includes/organisation-overview.html
```diff
@@ -1,3 +1,16 @@
+<table class="export-buttons">
+    <tr>
+        <td>
+        {% ifAsync ((organisationLevel === 'region') or (organisationLevel === 'ldu') or (organisationLevel === 'team')) %}
+        {% if ((authorisation===false) or (userRole==='Data Admin') or (userRole==='Manager') or (userRole==='System Admin')) %}
```
davidmck95: I don't think user role should affect reduction exports. All users should be able to export from region, ldu or team level

app/views/includes/organisation-overview.html
```diff
@@ -1,3 +1,16 @@
...
+        <a href="/probation/{{ organisationLevel }}/{{ linkId }}/{{ screen }}/reductions-csv" class="button reduction-export" style="float: right; margin-left:10px; margin-bottom: 10px">Export Reductions</a>
+        {% endif %}
+        {% endif %}
+        <a href="/probation/{{ organisationLevel }}/{{ linkId }}/{{ screen }}/caseload-csv" class="button sln-export" style="float: right; margin-bottom: 10px; ">Export Caseload</a>
```
davidmck95: Are export-caseload.html and export-reductions.html no longer used now? Should they be deleted or still used as {% include %}

app/assets/sass/application.scss
```diff
@@ -14,6 +14,13 @@ $path: "/public/images/";
   position: absolute !important;
 }
 
+a.reduction-export:visited{
+    color:white !important;
+}
+
+a.sln-export:visited{
+    color:white !important;
+}
```
davidmck95: If both of these classes do the same thing we should just use one css class.

app/views/includes/organisation-overview.html
```diff
@@ -1,3 +1,14 @@
+<table>
+    <tr>
+        <td>
+        {% ifAsync ((organisationLevel === 'region') or (organisationLevel === 'ldu') or (organisationLevel === 'team')) %}
+        <a href="/probation/{{ organisationLevel }}/{{ linkId }}/{{ screen }}/reductions-csv" class="button reduction-export" style="float: right; margin-left:10px; margin-bottom: 10px">Export Reductions</a>
+        {% endif %}
+        <a href="/probation/{{ organisationLevel }}/{{ linkId }}/{{ screen }}/caseload-csv" class="button sln-export" style="float: right; margin-bottom: 10px; ">Export Caseload</a>
```
davidmck95: Remove reduction-export and sln-export from class attributes if these are no longer specified. Same in export-caseload.html

c-mcateer: reduction-export and sln-export are used to tell the difference between the buttons on the e2e tests to make sure both are displayed

PR #219, opened 2017-11-28
Include region, ldu and team names and contracted hours in reduction …

davidmck95:
…exports


app/routes/overview.js
```diff
@@ -90,7 +90,9 @@ module.exports = function (router) {
     }
 
     return getReductionsExport(id, organisationLevel).then(function (result) {
+      console.log(result)
```
c-mcateer: Can the console.log be removed now that its working?

PR #225, opened 2017-11-29
fix case progress graph to allow it to display on firefox17

c-mcateer:
Replaced es6 arrow function so that the case progress graph would display in Firefox version 17


app/assets/javascripts/case-progress.js
```diff
@@ -16,7 +16,17 @@ caseProgress.forEach(function (caseInfo) {
   namesY.push(caseInfo.name)
 })
 
-maxLabelLength = Math.max(...(namesY.map(name => name.length)))
+function _toConsumableArray(arr) {
```
davidmck95: Indentation looks off, could be formatted better

PR #229, opened 2017-12-04
Fix/overdue terminations tag

c-mcateer:
Just changed the width of the case progress graph to allow all the graph hover labels to be seen!

### DIFF
--- a/app/views/case-progress.html
+++ b/app/views/case-progress.html
@@ -39,7 +39,9 @@
     </tbody>
 </table>
 
+<div class="graph">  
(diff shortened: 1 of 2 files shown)


app/views/case-progress.html
```diff
@@ -39,7 +39,9 @@
     </tbody>
 </table>
 
+<div class="graph">
```
davidmck95: What's the graph div used for?

c-mcateer: Oh I forgot I added that in, will removed it now and push it up again!

PR #233, opened 2017-12-05
Fix/capacity graph

davidmck95:
Y axis should increase automatically now instead of being hard coded to a max of 120.
Date searches which return no results will now display an error message.

### DIFF
--- a/app/views/capacity.html
+++ b/app/views/capacity.html
@@ -78,7 +78,7 @@
 </form>
 <br>
 {% if capacity.headings.length === 0 %}
-<span class="error-message">There is no data for this period.</span>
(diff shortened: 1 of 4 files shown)


app/views/capacity.html
```diff
@@ -75,8 +75,11 @@
     </div>
 
     <input id="caseload-filter-submit" type="submit" value="Filter" class="button">
-    
 </form>
+<br>
+{% if capacity.headings.length === 0 %}
+<span class="error-message">There is no data for this period.</span>
```
ddebella: Would it make sense to add the period in braces?

PR #242, opened 2017-12-11
Add a Totals row to the Inactive UPW table 

davidmck95:


test/unit/services/test-get-outstanding-reports.js
```diff
@@ -39,49 +46,129 @@ const LDU_RESULTS = [
     ow: 39,
     ot: 29,
     upw: 19,
-    t2aOw: 39,
-    t2aOt: 29,
-    t2aUpw: 19,
+    t2aOw: 0,
```
ddebella: Why put them to 0, shouldn't there be a value?

test/unit/services/test-get-outstanding-reports.js
```diff
@@ -39,49 +46,129 @@ const LDU_RESULTS = [
...
   {
-    linkId: 1,
-    name: 'Test ldu 1',
+    linkId: 2,
+    name: 'Test ldu 2',
     grade: 'Test grade 2',
     ow: 40,
     ot: 30,
     upw: 20,
-    t2aOw: 40,
```
ddebella: Same as above

test/unit/services/test-get-outstanding-reports.js
```diff
@@ -116,10 +203,10 @@ describe('services/get-outstanding-reports-view', function () {
   })
 
   it('should return a result object with outstanding reports for region', function () {
-    getOutstandingReportsView.resolves(LDU_RESULTS)
```
ddebella: 👍

PR #258, opened 2018-01-10
Fix/make headers float

c-mcateer:
Added in FloatThead to allow the table headers to always be seen as a user scrolls down a table


app/assets/javascripts/case-overview.js
```diff
@@ -7,8 +7,10 @@ $(document).ready(function() {
         "order": [[ 0, "asc" ]],
         columnDefs: [{
           targets: [],
-          orderable: false
+          orderable: false,
```
davidmck95: Last property in object columnDefs shouldn't have comma

app/assets/sass/_wmt/_wmt-data-table.scss
```diff
@@ -6,13 +6,30 @@
...
+    background-color:white
+}
+
+table.fixedHeader-floating.no-footer {
+    border-bottom-width:0
+}
+
+table.fixedHeader-locked {
+    position:absolute !important;
+    background-color:white
```
davidmck95: Missing semi-colons on line 11 and 20

PR #398, opened 2021-06-28
security upgrade, removed auth for existing AD

carlov20:


helm_deploy/wmt-web/values.yaml
```diff
@@ -26,7 +26,7 @@ generic-service:
     NODE_ENV: "production"
     WMT_DB_NAME: "hmpps_workload"
     WMT_DB_ARCHIVE_NAME: "hmpps_workload_archive"
-    WMT_WEB_AUTHENTICATION_ENABLED: "true"
+#    WMT_WEB_AUTHENTICATION_ENABLED: "true"
```
carlov20, line 29: wasn't sure if we wanted to keep this as a placeholder for when we do get to configuring auth or remove all together and start again when we do auth?

package.json
```diff
@@ -83,16 +83,16 @@
...
-    "sass": "^1.34.0",
-    "tedious": "^11.0.8",
+    "sass": "^1.35.1",
+    "tedious": "^11.0.9",
     "underscore": "^1.13.1",
-    "util": "^0.12.3",
+    "util": "^0.12.4",
     "validator": "^13.6.0",
-    "wmt-probation-rules": "ministryofjustice/wmt-probation-rules.git#v1.0.6"
+    "wmt-probation-rules": "github:ministryofjustice/wmt-probation-rules#v1.0.6"
```
markreesmoj, line 96: Does this mean we no longer need git installed on the docker image?

markreesmoj (approved):
👍

package.json
carlov20: no, still need git unfortunately. looks like this is a convention from npm (this happened when I ran npm upgrade)

PR #404, opened 2021-07-12
Wfp 325

carlov20:


.circleci/config.yml
```diff
@@ -106,6 +194,18 @@ workflows:
...
+          requires:
+            - build
+      - accessibility_test:
+          context: hmpps-common-vars
+          requires:
+            - build
+      - e2e_test:
+          context: hmpps-common-vars
+          requires:
+            - build
```
markreesmoj, line 208: In the event that we end up running all these tests in the pipeline, it's probably worth making this one depend on the accessibility-tests (or vice versa). This should avoid possible port collisions when running npm start and would mean we don't need to wait for the database to load twice

```suggestion
      - accessibility_test
```

app/views/capacity.html
```diff
@@ -305,8 +305,8 @@ <h3 class="govuk-heading-m">Inactive & UPW cases</h3>
   <table id="outstandingReports" class="govuk-table data-table dataTable">
     <thead>
       <tr>
-        <th colspan="2">{{ childOrganisationLevelDisplayText }}</th>
-        <th class="" colspan="5">Inactive & UPW cases</th>
+        <th colspan="2" id='organisationLevel'>{{ childOrganisationLevelDisplayText }}</th>
+        <th class="" colspan="5" id='inactiveHeader' >Inactive & UPW cases</th>
```
markreesmoj, line 309: HTML attributes should be double quoted I think. I suppose we could do with an HTML linter

app/views/reductions.html
```diff
@@ -101,7 +101,7 @@ <h2 id="headingScheduled" class="govuk-heading-l">Scheduled<span class="visually
 
     <h2 id="headingArchived" class="govuk-heading-l">Archived<span class="visuallyhidden"> reductions</span></h2>
 
-    <table class="govuk-table sln-table-reductions" id= "override-govuk-table">
```
markreesmoj, line 104: there is unfortunately some styling hanging off this id which we will need to check - see https://github.com/ministryofjustice/wmt-web/blob/master/app/assets/sass/application.scss#L190

PR #412, opened 2021-08-12
updated 404 error page content

liammcmurray:


liammcmurray:
@carlov20 hi, noticed you completed this task yesterday - but the content looks old compared to what was designed. Can you double check we have the latest version? Ta.

PR #438, opened 2021-10-20
WFP-685 stopped a system admin from being able to create or update to…

carlov20:
… a data admin

### DIFF
--- a/app/routes/user-rights.js
+++ b/app/routes/user-rights.js
@@ -161,26 +161,25 @@ const removeUserRole = function (username) {
 }
 
 const addUpdateUserRole = function (username, rights, loggedInUsername, fullname) {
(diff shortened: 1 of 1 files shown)


app/services/user-role-service.js
```diff
@@ -29,7 +30,7 @@ module.exports.addUserRole = function (userRole) {
 
 module.exports.getRole = function (role) {
   return getRole(role).then(function (result) {
-    return result
+    return result ?? { role: userRoles.STAFF }
```
markreesmoj, line 33: Nice - had to look this up as it appears to have landed since I last did any node
It returns the left hand side if it's not null, otherwise the right hand side
neater than 
`return result ? result : { role: userRoles.STAFF }`

PR #440, opened 2021-11-03
Wfp 351 google analytics

katie-cawthorne:
- I've added the `GOOGLE_ANALYTICS_KEY` value into the `hmpps-workload` secret in the namespaces


markreesmoj:
I really wanted to talk this through as a group rather than comment on a PR but I suppose we didn;t have time yesterday.

Do we have different keys for dev, preprod and prod? If not will all the traffic go into the same account? Or do we not set keys in dev and preprod?

PR #447, opened 2021-11-15
WFP-736 renamed data admin to super user

carlov20:


app/services/get-sub-nav.js
```diff
@@ -34,7 +34,7 @@ module.exports = function (id, organisationalUnitName, currentPath, workloadType
         navigation.push(new Link('NPS Caseload', baseLink + '/caseload'))
         navigation.push(new Link('CRC Caseload', baseLink + '/crc-caseload'))
         navigation.push(new Link('Case Progress', baseLink + '/case-progress'))
-        if (authorisation === false || userRole === 'Data Admin' || userRole === 'System Admin' || userRole === 'Manager') {
+        if (authorisation === false || userRole === 'Super User' || userRole === 'System Admin' || userRole === 'Manager') {
```
markreesmoj, line 37: ugh! both authorisation and userRole are passed into this method purely to write this horrible if statement! We have got a right job on our hands here. 
I appreciate the imperative not to rewrite while test coverage is not there and the scaffolding is so weak.
As a small first step towards improving it perhaps we could use the constants from app/constants/user-roles.js ?

app/views/admin.njk
```diff
@@ -7,7 +7,7 @@
   {% include "includes/title-subtitle.njk" %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-one-half">
-    {% if authorisation===false or userRole==='Data Admin'%}
```
markreesmoj, line 10: Similat to the above really, and I know this is not new code, but it would be great if we could reduce the number of places the string is hard coded. This might be better done in the controller?

app/views/archive-options.njk
```diff
@@ -7,7 +7,7 @@
   {% include "includes/title-subtitle.njk" %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-one-half">
-    {% if authorisation===false or userRole==='Data Admin'%}
+    {% if authorisation===false or userRole==='Super User'%}
```
markreesmoj, line 10: As above - can we replace the hard coded string with a constant?

app/views/export.njk
```diff
@@ -83,7 +83,7 @@
                             <label class="govuk-label govuk-radios__label" for="radio-inline-8">T2A</label>
                         </div>
                     </div>
-                    {% if authorisation===false or userRole==='System Admin' or userRole==='Data Admin' or userRole==='Manager'%}
+                    {% if authorisation===false or userRole==='System Admin' or userRole==='Super User' or userRole==='Manager'%}
```
markreesmoj, line 86: As above - can we replace the hard coded string with a constant?

app/views/includes/header.njk
```diff
@@ -20,13 +20,13 @@
       <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide navigation menu">Menu</button>
       <nav>
         <ul id="navigation" class="govuk-header__navigation " aria-label="Navigation menu">
-          {% if authorisation===false or userRole==='Data Admin' or userRole==='System Admin'%}
+          {% if authorisation===false or userRole==='Super User' or userRole==='System Admin'%}
```
markreesmoj, line 23: As above - can we replace the hard coded string with a constant?

app/views/includes/organisation-overview.njk
```diff
@@ -4,7 +4,7 @@
     <div class="govuk-grid-column-full">
 
             {% ifAsync ((organisationLevel === 'region') or (organisationLevel === 'ldu') or (organisationLevel === 'team')) %}
-                {% if authorisation===false or userRole==='System Admin' or userRole==='Data Admin' or userRole==='Manager'%}
+                {% if authorisation===false or userRole==='System Admin' or userRole==='Super User' or userRole==='Manager'%}
```
markreesmoj, line 7: I'm beginning to wish I hadn't started this

app/views/user-rights.njk
```diff
@@ -11,7 +11,7 @@
     {% set Manager=true %}
   {% elseif rights === 'System Admin' %}
     {% set SystemAdmin=true %}
-  {% elseif rights === 'Data Admin' %}
+  {% elseif rights === 'Super User' %}
     {% set DataAdmin=true %}
```
markreesmoj: should be `SuperUser`
actually, should be `superUser`
actually, should be done in the controller!

README.md
```diff
@@ -39,7 +39,7 @@ This application uses [HMPPS Auth](https://github.com/ministryofjustice/hmpps-au
 
 To seed data run `npm run seed-dev-data`, this will add users with elevated roles. To login to a user which has a specific role please use the following usernames:
 
-- WMT_DATA_ADMIN
+- WMT_SUPER_USER
```
markreesmoj, line 42: I think we will need to change hmpps-auth for this change to work when running locally against docker. You can currently log in as WMT_SUPER_USER (to my surprise) but you do not have admin privileges

I have since found you can log in locally as WMT_LITERALLY_ANYTHING as long as you use the correct password! Not clear on how or why this works yet.

markreesmoj (approved):
Usual nitpicking but at least it shows I read the whole PR

app/views/archive-options.njk
carlov20: This one is subject to change for WFP-678 so would rather clean this up in that story (this if statement shouldn't exist basically, is a mismatch from the rbac)

README.md
carlov20: we use wiremock to mock Delius https://github.com/ministryofjustice/wmt-web/blob/main/test/e2e/resources/wiremock/mappings/user-detail-success.json so currently anything beginning with WMT as a username will work. Then to give elevated roles we use https://github.com/ministryofjustice/wmt-web/blob/main/test/helpers/routes/authentication-helper.js#L2 to populate the database through the `seed-dev-data` mechanism

PR #456, opened 2021-11-23
WFP-730 added correct error when resolving breadcrumbs

carlov20:


app/services/get-breadcrumbs.js
```diff
@@ -27,37 +28,53 @@ const getBreadcrumbLevels = function () {
 
 const getOffenderManagerBreadcrumb = function (id, workloadType) {
   return getWorkloadOwnerOffenderManagerByWorkloadOwnerId(id).then(function (offenderManager) {
-    const breadcrumbs = [new Link(`${offenderManager[0].offender_manager_forename} ${offenderManager[0].offender_manager_surname}`, `/${workloadType}/offender-manager/${id}`)]
-    return getTeamBreadcrumb(offenderManager[0].team_id, workloadType).then(function (teamBreadcrumbs) {
-      return breadcrumbs.concat(teamBreadcrumbs)
-    })
+    if (offenderManager.length) {
```
carlov20, line 31: bit of a strange place to put all of this, this is the first time any level is retrieved, the alternative solution is to swap getBreadcrumbs everywhere in the routes to the last call in the chain and have a not found check there

test/e2e/view-internal-server.js
```diff
@@ -1,22 +0,0 @@
-const expect = require('chai').expect
```
markreesmoj, line 1: Are we unable to cause a 500 error by any other means? I would usually take that as a good sign!
I think we added this test because it was so easy to create a 500 this way

carlov20: I couldn't think of another known scenario for a 500 unfortunately

PR #742, opened 2022-09-29
Deadcode

carlov20:


test/integration/services/data/test-get-scenario.js
```diff
@@ -1,61 +0,0 @@
-const expect = require('chai').expect
```
markreesmoj, line 1: Why is this file being deleted? Looks like get-scenario isn't deleted in this PR

markreesmoj (changes requested):

test/integration/services/data/test-get-scenario.js
carlov20: I've reinstated it and moved the test only db queries into the test folder

PR #775, opened 2022-10-12
WFP-1194 change total / average cells to th

markreesmoj:


markreesmoj:
not sure this is the right way to go

PR #791, opened 2022-10-25
fix cms

carlov20:


app/services/data/get-cms-export.js
```diff
@@ -35,17 +35,17 @@ module.exports = function (id, type) {
...
-        result.contactCode = result.omContactCode
+      if (!result.contactcode) {
+        result.contactcode = result.omcontactcode
       }
-      if (!result.contactDate) {
-        result.contactDate = result.omContactDate
+      if (!result.contactdate) {
+        result.contactdate = result.omcontactdate
       }
     })
```
markreesmoj, line 50: Looks fine - let's see if it works. Worth noting that these names also appear here:

https://github.com/ministryofjustice/wmt-web/blob/main/app/services/get-export-csv.js#L37

markreesmoj (approved):

PR #795, opened 2022-10-26
WFP-1201: Update tab titles for Offender Management to help accessibility

sp-watson:


app/views/includes/page-title.njk
```diff
@@ -1,3 +1,11 @@
+{% if tabTitle %}
```
sp-watson, line 1: I hope this if statement will go away once we have done all menu options

markreesmoj:
Can we have some tests for this? Maybe the e2e tests can be adapted - they visit these pages?

markreesmoj (approved):
👍

PR #808, opened 2022-11-07
WFP-1496: Amend H1s for accessibility

sp-watson:


test/e2e/view-caseload.js
```diff
@@ -29,7 +29,7 @@ describe('View your caseload flow', () => {
       await browser.url(teamDefaultUrl + '/caseload')
       pageSubtitle = await $('.govuk-caption-xl')
       pageSubtitle = await pageSubtitle.getText()
-      expect(pageSubtitle).to.equal('Team')
+      expect(pageSubtitle).to.equal('Offender Manager')
```
sp-watson, line 32: Clearly the H1 was previously used to confirm the navigation went to the right page.

This causes us a problem now the H1 doesn't have that information. Looking at the HTML there is no obvious other way of finding out which organisational level each page is for - do we add something to every page to allow these tests to function?

If not, should we bother with these checks at all?

test/e2e/view-capacity-region.js
```diff
@@ -21,7 +21,7 @@ describe('View your caseload capacity flow', () => {
     await regionCasloadLink.click()
     pageSubtitle = await $('.govuk-caption-xl')
     pageSubtitle = await pageSubtitle.getText()
-    expect(pageSubtitle).to.equal('Region')
+    expect(pageSubtitle).to.equal('Offender Manager')
```
markreesmoj, line 24: This is the region caseload capacity view, based on what I can understand from the URL and the test filename

https://workload-measurement-preprod.hmpps.service.justice.gov.uk/probation/region/31/caseload-capacity

There isn't anything offender manager specific here

test/e2e/view-capacity-team.js
```diff
@@ -24,7 +24,7 @@ describe('Team', () => {
     it('should navigate to the team caseload capacity screen', async () => {
       pageSubtitle = await $('.govuk-caption-xl')
       pageSubtitle = await pageSubtitle.getText()
-      expect(pageSubtitle).to.equal('Team')
+      expect(pageSubtitle).to.equal('Offender Manager')
```
markreesmoj, line 27: As above

PR #816, opened 2022-11-08
WFP-1496: Update H1s for accessibility

sp-watson:


app/services/get-court-report-overview.js
```diff
@@ -13,7 +14,10 @@ module.exports = function (id, organisationLevel) {
         result.breadcrumbs = breadcrumbs
         result.overviewDetails = courtReportOverview
         result.title = result.breadcrumbs[0].title
-        result.subTitle = organisationalUnitType.displayText
+        if (organisationalUnitType.name === 'hmpps') {
+          result.title = organisationalUnitType.displayText
+        }
```
markreesmoj, line 19: It's a minor point but it might be nice to have this logic in one place

sp-watson: It's worth a look - I'll do that in the next PR

PR #829, opened 2022-11-11
moved creating the error list into the controller

carlov20:


test/unit/services/domain/test-capacity-date-range.js
```diff
@@ -44,180 +38,4 @@ describe('services/domain/capacity-date-range', function () {
     expect(capacityDateRange.toFields[1]).to.equal(VALID_TO_MONTH)
     expect(capacityDateRange.toFields[2]).to.equal(VALID_TO_YEAR)
   })
-
-  it('should throw ValidationError if a future date was given for capacity from date', function () {
```
carlov20, line 48: all of the actual validate functions are tested closer https://github.com/ministryofjustice/wmt-web/blob/error-list-in-controller/test/unit/services/validators/test-fieldset-validator.js, I didn't think it was worth duplicate tests testing the same underlying function

test/unit/services/domain/test-reduction.js
```diff
@@ -23,78 +23,4 @@ describe('services/domain/reduction', function () {
     expect(reduction.status).to.be.equal(reductionStatusType.ACTIVE)
     expect(reduction.reductionSubmitter).to.equal(150)
   })
-
-  it('should raise a ValidationError if Hours is not supplied', function () {
```
carlov20, line 27: all of the actual validate functions are tested closer https://github.com/ministryofjustice/wmt-web/blob/error-list-in-controller/test/unit/services/validators/test-fieldset-validator.js or https://github.com/ministryofjustice/wmt-web/blob/error-list-in-controller/test/unit/services/validators/test-field-validator.js, I didn't think it was worth duplicate tests testing the same underlying function

markreesmoj (approved):
🧹

PR #874, opened 2022-12-14
WFP-1696 replace moment with Date on client side

markreesmoj:
https://dsdmoj.atlassian.net/browse/WFP-1696


app/assets/javascripts/archive-reductions-new.js
```diff
@@ -1,19 +1,4 @@
 var search
```
markreesmoj, line 1: No sortable headers on this table

carlov20 (approved):
🧹

PR #889, opened 2022-12-29
WFP-1197: Amend graph line colours for DAC report

sp-watson:
I spent quite a while trying to uncover a way of changing the plot line (or background) colours and couldn't find anything.

In the end I found the definition of the line colours in the Plotly JS file and changed them. This is bad because we will lose those changes if we ever upgrade, but we should only upgrade if we can set the line and/or background colours.

For reference the colours I changed were:

 - Blue line from 1f77b4 to 1b6aa1
 - Orange line from ff7f0e to db6600

i did consider writing a test that will fail if we upgrade - maybe we should?


carlov20 (approved):
🎨

PR #892, opened 2023-01-03
WFP-1195: Improve accessibility of red text

sp-watson:
Had to do 3 things:

 - Add text above every table that might have red text, which explains what the red text means 
 - Made the red text bold
 - Added screen-reader only text for each cell that has red text

### DIFF
--- a/app/assets/sass/application.scss
+++ b/app/assets/sass/application.scss
@@ -48,6 +48,11 @@ tr.shown td.details-control {
   background: url('../images/details_close.png') no-repeat center center;
 }
 
+.capacity-red {
+  color: #942514 !important;
+  font-weight: bold;
+}
+
 .width800 {
   width: 800px;
 }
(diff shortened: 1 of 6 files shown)


test/e2e/view-overview-om.js
```diff
@@ -30,7 +30,8 @@ describe('Offender Manager', function () {
 
       let capacity = await $('.sln-capacity')
       capacity = await capacity.getText()
-      expect(capacity).to.equal(`${workloadOwnerCapacity}%`)
+      const capacityPercent = capacity.substring(0, capacity.indexOf('%'))
```
sp-watson, line 33: Needed to extract the text before the percent, otherwise the hidden text would be included

app/assets/sass/application.scss
```diff
@@ -48,6 +48,11 @@ tr.shown td.details-control {
   background: url('../images/details_close.png') no-repeat center center;
 }
 
+.capacity-red {
```
sp-watson, line 51: Moved this style out of the table styles as it is now used in the text above the table

carlov20 (approved):
💬

PR #894, opened 2023-01-05
WFP-1194: Improve accessibility of table footers and summary rows

sp-watson:


app/assets/sass/_wmt/_wmt-data-table.scss
```diff
@@ -61,6 +61,18 @@ table.data-table tfoot tr {
     background: rgb(199, 192, 192);
 }
 
+table.data-table tfoot tr th {
```
sp-watson, line 64: This is to make these summary rows look like they currently does

app/assets/sass/_wmt/_wmt-data-table.scss
```diff
@@ -61,6 +61,18 @@ table.data-table tfoot tr {
...
+    background: rgb(199, 192, 192);
+    font-size: 16px;
+    font-weight: normal;
+}
+
+table.data-table tfoot tr th:first-child {
```
sp-watson, line 70: This is to style the left-most column of the summary row to match the current style

carlov20 (approved):

PR #931, opened 2023-02-21
WFP-1840 amended permissions for exports

carlov20:


sp-watson (approved):
I am assuming the changes to access were agreed with Pauline?

PR #1020, opened 2023-08-29
Wfp 2085

wood-n:


AndySweilam:
It is a bit different than what we did in WMT-Worker. Let's catch up tomorrow about it 👨‍💻

PR #1029, opened 2023-12-15
Revert migrate auth endpoints

gregkhawkins:


AndySweilam (approved):
LGTM, let's hope we don't need to use it

PR #1134, opened 2025-05-22
Update node to LTS 22.14 and fix e2e tests

katie-cawthorne:
So this PR is very lengthy, I apologise for the amount of commits but I had an issue where the tests were fixed locally but failed in the pipeline, so was very hard to know that it was fixed without committing changes. The changes are:

1. I updated nodeJS to LTS 22.14, this was to fix a security vulnerability. When the post commit build script ran the e2e 
tests were failing so most of the commits are fixing those tests.

2. A common theme with the tests was around timeout issues caused by the browser not behaving as fast as expected. WebdriverIO waits for something that never appears. I created browser helper functions to add some timeout when navigating to a page and clicking a button. I also needed to use the waitForDisplayed function before accessing some elements in individual tests to fix some failing tests.

3. Another issue I came across was around the bidi protocol which is part of webdriverIO version 9, many tests were failing with bidi timeout errors. Please see this for more info https://github.com/webdriverio/webdriverio/issues/13909. I updated the wdio config to use the classic protocol `wdio:enforceWebDriverClassic': true` and this seemed to fix the errors.

4. There were some tests around adding a reduction that were failing as the end dates used were supposed to be in the future but they were now in the past so this was causing the reduction to display as an 'archived' reduction instead of a current reduction.

5. There was an issue with some tests that were using a hardcoded `data-select2-id` to access text used in a dropdown. The data-select2 is dynamic so this was why it couldn't find the element. I changed the test to find the element by the text instead. 

6. There was also an issue getting data from the user preference service and the count of unallocated cases around accessing the JSON object so I needed to configure this.


sambailey-moj (approved):
Looks good. The changes (there are a lot of them) are very clean & consistent

package.json
```diff
@@ -108,6 +108,7 @@
     "json2csv": "5.0.7",
     "jwt-decode": "4.0.0",
     "knex": "3.1.0",
+    "latest": "^0.2.0",
```
iainjchambers-moj, line 111: Just wondering where this came from? Can't see any other references to it and I'm wondering if it's a result of a typo since it seems so random?

iainjchambers-moj (commented):
Awesome work. One very minor nitpick

GaReid (approved):

PR #1136, opened 2025-06-03
Point tests to wiremocks

iainjchambers-moj:


README.md
```diff
@@ -150,6 +150,10 @@ Run tests using the following commands:
 ```
 docker compose up -d
 ```
+- post mappings to wiremock
+```
+find test/e2e/resources/wiremock/mappings/*.json -exec curl --request POST --url http://localhost:8099/__admin/mappings --header 'content-type: application/json' --data-binary "@{}" \;
```
sambailey-moj, line 155: My only question, is could this be wrapped in an node command? Happy with this implementation, and don't want to hold this up - might be one to implement in a future ticket if it's a bit of work?

sambailey-moj (dismissed):
Looks great

PR #1143, opened 2025-06-19
fixing unallocated cases notifications

sambailey-moj:


test/e2e/view-overview-national.js
```diff
@@ -43,7 +43,7 @@ describe('National', function () {
       const allocationsNumber = await $('#notifications')
       await allocationsNumber.waitForDisplayed({ timeout: 30000 })
       const number = await allocationsNumber.getText()
-      return expect(number).to.equal('42')
+      return expect(number).to.equal('+42')
```
iainjchambers-moj, line 46: this shouldn't be happening as far as I'm aware, the `+` symbol should only be put there by line 20 on `getUnallocatedCaseCount.js` as a fail safe for when the allocation service throws a 500, otherwise it should just be a number

iainjchambers-moj:
LGTM 👍

